PR #87, opened 2021-12-09
Fix bug in JamEvent::isEmpty

svenssonaxel:
It is common for an empty JamEvent to be marked as the last event
before the execution of the next incoming signal. In these cases,
JamEvent::isEmpty used to erroneously return false, causing the Jam
event log to contain lines for file "unknown_file_32767", line 65535.


svenssonaxel:
Fixes #86

PR #154, opened 2022-08-03
Rondb 151: Add REST API Server

smkniazi:


mronstro (approved):
Consider putting short functions as inline in header file (done in some places)

PR #172, opened 2022-11-02
RONDB-180: Fix cluster reconnect for unload and setpartition operations

smkniazi:


mronstro (approved):
Looks good, one minor comment, there is no check if the error code is UnknownErrorResult, at the moment I think it is the only one, but could be that in the future other error codes emerge. So could be good to handle that now.

smkniazi:
UnkonwnErrorResults is handled https://github.com/logicalclocks/rondb/blob/a350044374a412883f026ab8049b1f76485c1faa/storage/ndb/clusterj/clusterj-core/src/main/java/com/mysql/clusterj/core/SessionImpl.java#L1791

PR #176, opened 2022-11-07
RONDB - 184: Docker Build Changes

olapiv:
* Added Dockerfile with base image ubuntu:22.04 for experimental arm64 builds
* Using caching of downloads & builds within Dockerfiles
* Bumped sysbench & dbt2 versions to accommodate for arm building (`build_bench.sh`)
* Dynamic naming of tarballs depending on building architecture (`create_rondb_tarball.sh`)
* Removed `docker-build.sh`; Logic is now in Dockerfile
* Formatting of scripts

TODO:
* Make same PR to other RonDB versions


build_scripts/release_scripts/create_rondb_tarball.sh
```diff
@@ -7,21 +7,54 @@ OUTPUT_DIR_ABS=$3
...
+echo "WARNING: No solution figure out how to figure out OS variant"
+if [ "$CPU_ARCH" == "amd64" ]; then
+  CPU_ARCH_FULL="x86_64"
+elif [ "$CPU_ARCH" == "arm64" ]; then
+  echo "WARNING: Guessing that we are running with arm64_v8"
+  CPU_ARCH_FULL="arm64_v8"
+else
+  echo "Cannot figure out the exact CPU architecture"
+  exit 1
+fi
```
olapiv: @mronstro I'm not very happy about the logic I added above here, but do you have any better suggestions?

Dockerfile
```diff
@@ -0,0 +1,120 @@
...
+ENV BOOST_LABEL="boost_${BOOST_V_UNDERSCORE}"
+ENV BOOST_TARBALL="${BOOST_LABEL}.tar.gz"
+RUN wget --progress=bar:force https://boostorg.jfrog.io/artifactory/main/release/${BOOST_VERSION}/source/${BOOST_TARBALL} \
+    && tar xzf ${BOOST_TARBALL} \
+    && rm ${BOOST_TARBALL}
+# Important env variable for RonDB compilation
+ENV BOOST_ROOT="/${BOOST_LABEL}"
+
+# Ubuntu 22.04 contains openssl-3.x
+# RonDB requires openssl-1.1.1m and will look for it in $OPENSSL_ROOT
```
mronstro: We should be on the latest openssl 1.1.1, this is now openssl 1.1.1s

Dockerfile.oraclelinux7
```diff
@@ -1,108 +1,167 @@
...
+ARG UID=1000
+ARG GID=1000
+
+FROM oraclelinux:7-slim as rondb-build-dependencies
+
+ARG USER
+ARG UID
+ARG GID
+ARG CMAKE_VERSION=3.23.2
+ARG OPEN_SSL_VERSION=1.1.1o
```
mronstro: Now openssl is at 1.1.1s

build_scripts/release_scripts/Jenkinsfile
```diff
@@ -48,12 +49,18 @@ pipeline {
...
 
-                release=""
+                RELEASE_ARG=
                 if [ "$RELEASE_BUILD" = true ]; then
-                  release=" -r "
+                  RELEASE_ARG="--build-arg RELEASE_TARBALL=$RELEASE_TARBALL"
                 fi
 
-                ./docker-build.sh -s ../.. -o output -j ${BUILD_CORES} -d $release
+                docker buildx build . \
```
smkniazi, line 62: Add above 
export BUILDKIT_PROGRESS=plain

Dockerfile.oraclelinux7
```diff
@@ -0,0 +1,164 @@
+# syntax=docker/dockerfile:1.3-labs
+
+# This Dockerfile is used for x86 production builds. Oracle Linux 7 comes with glibc 2.17,
+# which is supported by all Linux distributions used by Hopsworks: CentOS >=7 and Ubuntu >=16.
+# Glibc tends to be backwards compatible, but not forwards compatible.
+# Using this file for arm64 requires too many modifications.
+
+ARG USER=mysql
+ARG UID=1000
```
smkniazi: you need set UID & GID in the Jenkins file

Dockerfile.oraclelinux7
```diff
@@ -0,0 +1,164 @@
...
+ARG GID=1000
+
+FROM oraclelinux:7-slim as rondb-build-dependencies
+
+ARG USER
+ARG UID
```
smkniazi: why do we have to redefine UID / GID here? Is it dockerx thing?

Dockerfile.oraclelinux7
```diff
@@ -0,0 +1,164 @@
...
+RUN --mount=type=cache,target=$DOWNLOADS_CACHE_DIR \
+    wget -N --progress=bar:force -P $DOWNLOADS_CACHE_DIR \
+        https://boostorg.jfrog.io/artifactory/main/release/${BOOST_VERSION}/source/${BOOST_TARBALL} \
+    && tar xzf $DOWNLOADS_CACHE_DIR/${BOOST_TARBALL} -C .
+# Important env variable for RonDB compilation
+ENV BOOST_ROOT="/${BOOST_LABEL}"
+
+# RonDB 21.04 requires OpenSSL version 1.1.1m and will look for 
+# it in $OPENSSL_ROOT. We will not overwrite the default OpenSSL, but 
+# just install the specified version.
```
smkniazi: " We will not overwrite the default OpenSSL, but just install the specified version."
?? aren't you calling make install? 
Perhaps you want to say that you are not changing the system's openssl lib

Dockerfile.oraclelinux7
```diff
@@ -0,0 +1,164 @@
...
+
+ARG DEPLOY_TO_REPO
+ENV DEPLOY_ARG=${DEPLOY_TO_REPO:+-d}
+
+ARG RELEASE_TARBALL
+ENV RELEASE_ARG=${RELEASE_TARBALL:+-r}
+
+RUN mkdir rondb-src rondb-bin rondb-tarball
+RUN --mount=type=bind,source=.,target=rondb-src \
+    --mount=type=cache,target=rondb-bin,uid=$UID,gid=$GID,id=oracle7-rondb2104-bin \
```
smkniazi: We need to re-discuss this as this change make's it very hard to use docker build in a development environment

Dockerfile.ubuntu22
```diff
@@ -0,0 +1,147 @@
...
+    && make install \
+    && cd .. \
+    && rm -r cmake-$CMAKE_VERSION
+
+RUN groupadd $USER --gid $GID && adduser $USER --uid $UID --ingroup $USER --shell /bin/bash
+
+WORKDIR /home/$USER
+
+# For the deployment of Java artifacts to the Maven repository
+COPY <<-"EOF" .m2/settings.xml
```
smkniazi: For ubuntu22 we only have to deploy that RonDB tarball, Java artifacts should be skipped. Change the deploy.sh accordingly

build_scripts/release_scripts/Dockerfile
```diff
@@ -1,108 +0,0 @@
-# See docker-build.sh for usage
```
smkniazi, line 1: next time use `git mv ...` to move files around instead of deleting files in on dir and then creating the same file somewhere else. This messes up git history, and git blame

build_scripts/release_scripts/Jenkinsfile
```diff
@@ -36,9 +36,10 @@ pipeline {
             steps {
```
smkniazi, line 36: I do not see changes for Docker.ubuntu22.

smkniazi: All caches should be dropped. It has to be a clean build

build_scripts/release_scripts/build_all.sh
```diff
@@ -18,105 +17,122 @@ if [[ -z "${BOOST_ROOT}" ]]; then
...
+        Path to output directory where the build process will copy the final tarball
+-j=build_threads
+        No of build threads. This is passed to $(make -j$build_threads)
+-d      Deploy to remote repo.hops.works
+-r      Make release tarballs. Takes longer
+EOF
 }
 
+# Defaults
+CORES=$(($(nproc) / 2 + 1))
```
smkniazi: get rid of this, this has to come from -j switch

build_scripts/release_scripts/build_all.sh
```diff
@@ -18,105 +17,122 @@ if [[ -z "${BOOST_ROOT}" ]]; then
...
+    exit 0
+    ;;
+  esac
 done
-shift $((OPTIND-1))
+shift $((OPTIND - 1))
 [ "$1" = "--" ] && shift
 
-
 if [[ "$SRC_DIR" == "" ]]; then
```
smkniazi, line 77: why did you remove this check? build_all.sh can also be called outside docker.

build_scripts/release_scripts/build_all.sh
```diff
@@ -136,16 +152,14 @@ echo "_____________ BUILDING BENCHMARKS _____________"
...
-echo "_____________ Building Tarball _____________"
+echo "_____________ BUILDING TARBALL _____________"
 cd $TEMP_BUILD_DIR_ABS
 $SRC_DIR_ABS/build_scripts/release_scripts/create_rondb_tarball.sh $RONDB_VERSION $TEMP_BUILD_DIR_ABS $OUTPUT_DIR_ABS
 
 if [ "$DEPLOY" = true ]; then
-  echo "_____________ Deploying Tarball _____________"
-  cp $SRC_DIR_ABS/build_scripts/release_scripts/id_rsa $TEMP_BUILD_DIR_ABS
+  echo "_____________ DEPLOYING TARBALL _____________"
+  cp $SRC_DIR_ABS/id_rsa $TEMP_BUILD_DIR_ABS
```
smkniazi: Path looks bit odd, need to test it on Jenkins

build_scripts/release_scripts/build_all.sh
```diff
@@ -136,16 +152,14 @@ echo "_____________ BUILDING BENCHMARKS _____________"
...
   cd $TEMP_BUILD_DIR_ABS
   chmod 600 id_rsa
```
smkniazi: I expect this to fail on on Jenkins if you do not set the UID and GID for the Jenkins user

build_scripts/release_scripts/create_rondb_tarball.sh
```diff
@@ -7,21 +7,61 @@ OUTPUT_DIR_ABS=$3
 
 cd $TEMP_BUILD_DIR_ABS
 
+# The following code of how to determine OS and CPU architecture is not optimal.
+# Essentially, we are trying to keep the naming conventions on repo.hops.works
+#   Examples:
+#   rondb-21.04.2-linux-glibc2.17-x86_64.tar.gz
+#   rondb-21.04.2-macos-xcode-13.1-arm64_v8.tar.gz
```
smkniazi: xcode? it is an IDE

build_scripts/release_scripts/create_rondb_tarball.sh
```diff
@@ -7,21 +7,61 @@ OUTPUT_DIR_ABS=$3
...
+CPU_ARCH=$(uname -m)
+
+if [ "$CPU_ARCH" == "amd64" -o "$CPU_ARCH" == "x86_64" ]; then
+  CPU_ARCH_FULL="x86_64"
+elif [ "$CPU_ARCH" == "arm64" -o "$CPU_ARCH" == "aarch64" ]; then
+  # ARMv8 introduced 64-bit
+  CPU_ARCH_FULL="arm64_v8"
+else
+  echo "Cannot figure out the exact CPU architecture"
+  CPU_ARCH_FULL="unknown_arch"
```
smkniazi: throw hard error

build_scripts/release_scripts/deploy.sh
```diff
@@ -31,20 +31,18 @@ if [[ ! -f "$JAR_FILE" ]]; then
 fi
 
 mvn deploy:deploy-file -Dfile=$JAR_FILE -DgroupId=com.mysql.ndb -DartifactId=clusterj-rondb \
```
smkniazi, line 42: this file is only for x86_64. add a check so that it is not called for any other architecture

build_scripts/release_scripts/docker-build.sh
```diff
@@ -1,134 +0,0 @@
-#!/bin/bash
```
smkniazi: You have to add a small script in the root with instructions on how to start building. It is cumbersome to look into the Dockerfile to see what parameters are supported. I had to extract the build command from the Jenkins file

smkniazi (changes requested):
Good work, this will greatly improve the deployment
My main concerns are
1. cache mount points, isn't there a workaround to export the build folder? 
2. You have to extend the Jenkins file to support building for Docker.ubuntu22
3. Testing on Jenkins

Dockerfile.oraclelinux7
```diff
@@ -0,0 +1,164 @@
...
+    source scl_source enable devtoolset-9 \
+    && rondb-src/build_scripts/release_scripts/build_all.sh \
+    -s rondb-src \
+    -b rondb-bin \
+    -o rondb-tarball \
+    -j $THREADS_ARG \
+    $DEPLOY_ARG $RELEASE_ARG
+
+# run with --output <output-folder>
```
smkniazi, line 149:
```
export BUILDKIT_PROGRESS=plain
docker buildx build . \
-f Dockerfile.oraclelinux7 \
--tag rondb-build-all:21.04.9 \
--target build-all \
--build-arg BUILD_THREADS=6 --output /tmp/

```

failed on my machine 

```
#27 exporting to client
#27 copying files
#27 copying files 307.51MB 5.0s
#27 copying files 643.92MB 10.0s
#27 copying files 654.41MB 10.3s done
#27 ERROR: error from receiver: failed to create device /tmp/dev/console: operation not permitted
------
 > exporting to client:
------
error: failed to solve: error from receiver: failed to create device /tmp/dev/console: operation not permitted
```

smkniazi (changes requested):

Dockerfile.oraclelinux7
olapiv: It's a Docker multi-stage building thing: https://stackoverflow.com/a/53683625/9068781

Dockerfile.oraclelinux7
olapiv: Yeah that's what I wanted to say

build_scripts/release_scripts/create_rondb_tarball.sh
olapiv: Xcode comes with all the C/C++ libraries in Mac. It's more than an IDE.

Dockerfile.ubuntu22
olapiv: Why should the Java artefacts be skipped there? And also, so far it is just meant for private use anyways, since we're not planning to deploy 21.04 on ARM. So I didn't think that I'd add it to the Jenkinsfile in any case.

In other words, we wouldn't only skip deploying Java artefacts, but everything.
So why should specifically Java artefacts be skipped?

build_scripts/release_scripts/Dockerfile
olapiv: Well that's because my first draft only had the Ubuntu Dockerfile, which I wrote from scratch. Until I was told that we need to keep the Oracle Linux one again, so I brought it back to life.

build_scripts/release_scripts/Jenkinsfile
olapiv: Added a --no-cache flag.

Regarding ubuntu22 - we're not deploying RonDB 21.04.9 for ARM? Since we agreed to just do that for 22.10..? But this file will be the basis of that. This file is for the case where Mikael / me want to build something for personal usage.

@mikaelronstrom, had I understood that correctly?

build_scripts/release_scripts/build_all.sh
olapiv: It's just beneath. Now it's an early return and the `SRC_DIR_ABS` variable is not defined within an if statement. Not crucial in bash of course.

build_scripts/release_scripts/deploy.sh
olapiv: See my other comment regarding the Jar files (i.e. why Java only for x86?)

Dockerfile.oraclelinux7
olapiv: The `--output` flag is only meant to be used for `--target get-package-all`. Can you try again?

build_scripts/release_scripts/docker-build.sh
olapiv: I would rather adjust the documentation / add a comment to the README? @mikaelronstrom, what's your opinion?

build_scripts/release_scripts/build_all.sh
olapiv: How come?

Dockerfile.oraclelinux7
olapiv: Why?

Dockerfile.oraclelinux7
olapiv: Added docker-build.sh in the release scripts again. Could not find a way to let the build process write into a local directory (not counting a copy after things have finished).

olapiv: Just for the record - doing it with `docker run` and a mounted directory seems to be taking a lot longer than within `docker build`.

Dockerfile.oraclelinux7
smkniazi: Jenkins user ID and GID is not 1000

Dockerfile.ubuntu22
smkniazi: Java artifacts will have the same version. We do not want to deploy jars multiple times with the same version.

build_scripts/release_scripts/build_all.sh
smkniazi: As mentioned earlier, the uid and gid are different on JEnkins

build_scripts/release_scripts/deploy.sh
smkniazi: Java is platform independent

smkniazi (changes requested):
There is not much to fix, 
Do the following before we merge it in the tree
1. Link this PR with JIRA issue
2. Test on Jenkins

build_scripts/release_scripts/build_all.sh
olapiv: The user on my laptop is vincent/501 and it works. The Jenkinsfile still uses the mounted build cache within the Dockerfile, which I already set to the default UID & GID as specified in the following line:

`--mount=type=cache,target=rondb-bin,uid=$UID,gid=$GID,id=oracle7-rondb2104-bin`

build_scripts/release_scripts/deploy.sh
olapiv: Have a look now!

Dockerfile.ubuntu22
olapiv: I'd like to keep this tbh, just because it becomes difficult to back-track what is missing if we want to deploy from here again for some reason. Someone else might think that file is something specific to Oraclelinux. It's going to do an early return in the deploy.sh though. I also added a comment.

smkniazi (approved):

PR #233, opened 2023-03-07
RONDB-248: Fix CLI arguments

olapiv:
* The default configuration needs to be valid, for the rdrs to work as a binary.
* Also created a default config file for Docker tests
* The `resource` directory will be used in later PRs

### DIFF
--- a/storage/ndb/rest-server/rest-api-server/Makefile
+++ b/storage/ndb/rest-server/rest-api-server/Makefile
@@ -19,6 +19,7 @@ HOSTNAME=`hostname`
 BRANCH=`git rev-parse --abbrev-ref HEAD`
 SERVER_BIN="rdrs"
 RDRS_LIB_DIR?=./../data-access-rondb/build
+RDRS_CONFIG_FILE=$(PWD)/resources/config/config_dockertest.json
 OUTPUT_DIR?=./bin/server/${SERVER_BIN}
 
 COVERAGE_OUTPUT_FILE?=./coverage.out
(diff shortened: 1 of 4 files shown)


storage/ndb/rest-server/rest-api-server/internal/config/init_defaults.go
```diff
@@ -27,6 +27,15 @@ import (
 var globalConfig AllConfigs
 var mutex sync.Mutex
 
+/*
```
mronstro, line 30: Not sure I followed the logic in the comment, I interpret it as
if no ENV we don't check CLI argument

storage/ndb/rest-server/rest-api-server/Makefile
```diff
@@ -60,7 +61,7 @@ build_proto:
 test:
 	@go clean -testcache
 	@echo "Getting cross-package (integration) coverage and writing it to $(COVERAGE_OUTPUT_FILE)"
-	CGO_LDFLAGS="-g -O3 -L${RDRS_LIB_DIR} -lrdrclient" go test -p 1 -coverprofile $(COVERAGE_OUTPUT_FILE) -coverpkg ./... `go list ./... | grep -v cmd`
+	CGO_LDFLAGS="-g -O3 -L${RDRS_LIB_DIR} -lrdrclient" RDRS_CONFIG_FILE=$(RDRS_CONFIG_FILE) go test -p 1 -coverprofile $(COVERAGE_OUTPUT_FILE) -coverpkg ./... `go list ./... | grep -v cmd`
```
smkniazi, line 64: With this change I will not be able to run tests as I do not use docker

smkniazi (approved):

storage/ndb/rest-server/rest-api-server/Makefile
olapiv: No, the CLI config overwrites the ENV config

mronstro (approved):

PR #235, opened 2023-03-07
RonDB-248: More Config Validation

olapiv:


mronstro (approved):
Check port number not too big also

olapiv:
It's uint16

PR #241, opened 2023-03-13
RONDB-286

smkniazi:


storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_test.go
```diff
@@ -46,19 +46,12 @@ func TestAPIKey(t *testing.T) {
 	}
 	defer dal.ShutdownConnection()
 
-	databases := []string{testdbs.DB001, testdbs.DB002}
-	err, dropDatabases := testutils.CreateDatabases(true, databases...)
-	if err != nil {
-		t.Fatalf("failed creating databases %v; error: %v ", databases, err)
-	}
-	defer dropDatabases()
```
olapiv, line 54: I personally sometimes run these tests by themselves in my IDE. Can you add them again with/without the sentinel check? Since creating these few dbs is not a big overhead anyways.

storage/ndb/rest-server/rest-api-server/internal/integrationtests/initialiase_tests.go
```diff
@@ -40,15 +39,21 @@ func InitialiseTesting(conf config.AllConfigs, createOnlyTheseDBs ...string) (cl
...
 		dbsToCreate = testdbs.GetAllDBs()
 	}
-	err, dropDatabases := testutils.CreateDatabases(conf.Security.UseHopsworksAPIKeys, dbsToCreate...)
-	if err != nil {
-		cleanupTLSCerts()
-		return cleanup, fmt.Errorf("failed creating databases; error: %v", err)
+
+	//creating databases for each test run is very slow
+	//if sentinel DB exists then skip creating DBs
+	//drop the "sentinel" DB if you wan to recreate all the databasese.
```
olapiv: * want
* databases

storage/ndb/rest-server/rest-api-server/resources/testdbs/embeddings.go
```diff
@@ -147,3 +147,10 @@ const DB023 = "db023"
 var DB024Scheme string
 
 const DB024 = "db024"
+
+//This is sentinel DB
+//If this exists then we have Successfully initialized all the DBs
```
olapiv: successfully

storage/ndb/rest-server/rest-api-server/internal/testutils/schemata.go
```diff
@@ -42,7 +42,7 @@ func CreateDatabases(
 		err = runQueries(createSchema, dbConn)
 		if err != nil {
 			cleanupDbsWrapper(dropDatabases)()
-			err = fmt.Errorf("failed running createSchema for db '%s'; error: %w", db, err)
+			err = fmt.Errorf("failed running createSchema for db '%s'; error: %v", db, err)
```
olapiv: Go has a special directive for errors when formatting strings `%w`. You can basically unwrap the errors later layer by layer if you use it.

storage/ndb/rest-server/rest-api-server/internal/integrationtests/utils.go
```diff
@@ -478,7 +478,7 @@ func sendGRPCBatchRequest(t *testing.T, testInfo api.BatchOperationTestInfo) (in
 
 	// Create gRPC client
 	conf := config.GetAll()
-	conn, err := testutils.CreateGrpcConn(t, conf.Security.UseHopsworksAPIKeys, conf.Security.RequireAndVerifyClientCert)
+	conn, err := testutils.CreateGrpcConn(t, conf.Security.UseHopsworksAPIKeys, conf.Security.EnableTLS)
```
olapiv, line 481: This removes the test case where TLS is enabled, but the client doesn't use it. The config validation stops you from setting `RequireAndVerifyClientCert=true` && `conf.Security.EnableTLS=false` if that's what you're worried about.

olapiv (changes requested):
Reviewed the GO & mtr part.
@mronstro, you can have a go at the rest :)

storage/ndb/rest-server/rest-api-server/internal/integrationtests/initialiase_tests.go
smkniazi: fixed

storage/ndb/rest-server/rest-api-server/resources/testdbs/embeddings.go
smkniazi: fixed

storage/ndb/rest-server/rest-api-server/internal/testutils/schemata.go
smkniazi: reverted

storage/ndb/rest-server/rest-api-server/internal/integrationtests/utils.go
smkniazi: CreateGrpcConn has the following signature 
```
func CreateGrpcConn(t testing.TB, withAuth, withTLS bool) (*grpc.ClientConn, error) {
```

It does not have a param for RequireAndVerifyClientCert. if you run the tests such that you set the RequireAndVerifyClientCert=false in the config file then all the grpc tests will fail without this fix

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_test.go
smkniazi: Right now one has to wait for 30 sec each time running a single PK or Batch test. So I changed those tests to create the databases only once at the start. 
I changed this test also to create all databases 1) to be consistent with other tests that need to access databases 2)  hops works database breaks if you do not create all the databases and then some tests will fail with authorization exception. For now, it is fine.

olapiv: But right now, this test isn't creating any database at all? So if I run this test by itself it will fail?

smkniazi: you are right. I forgot to commit a new file. I have pushed "internal/security/apikey/wrapper_test.go"

storage/ndb/rest-server/rest-api-server/internal/integrationtests/utils.go
olapiv: Why do they fail? I don't think they should?

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_test.go
olapiv: Okay cool. Still though - this means if I want to run this test by itself, it will take 7s longer than without this change. That's okay, but seems a bit unnecessary. I can agree with all other changes, even though I personally prefer running against a Docker cluster, so I don't have the cleanup.

storage/ndb/rest-server/rest-api-server/internal/integrationtests/utils.go
smkniazi: Please see the above comment. you were passing RequireAndVerifyClientCert to EnableTLS param.

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_test.go
smkniazi: if you run any test then there will be a fixed cost at the start for creating all the databases. you have to do this only once until there is a change in the schema.

storage/ndb/rest-server/rest-api-server/internal/integrationtests/utils.go
olapiv: Yeah, but from the client side. So if `RequireAndVerifyClientCert=true`, then `EnableTLS=true` has to be true on the server side.  If `RequireAndVerifyClientCert=false`, the server can be running both `EnableTLS=true | false`, so we can quickly test that as well. Or am I misunderstanding something about TLS?

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_test.go
olapiv: Yeah I get it, it's just 1s (create 3 dbs) vs. 7s (create 20 dbs) in this case

storage/ndb/rest-server/rest-api-server/internal/integrationtests/utils.go
olapiv: And so I'm just curious why it would fail?

storage/ndb/rest-server/data-access-rondb/src/db-operations/pk/common.cpp
```diff
@@ -338,9 +318,10 @@ RS_Status SetOperationPKCol(const NdbDictionary::Column *col, NdbOperation *oper
     }
 
     // operation->equal expects a zero-padded char string
-    char pk[MAX_KEY_SIZE_IN_WORDS*4];
-    std::fill (pk, pk + col->getLength(), 0);
-    
+    char pk[CHAR_MAX_SIZE_IN_BYTES];
+    assert(col->getLength() <= CHAR_MAX_SIZE_IN_BYTES);
+    std::fill(pk, pk + col->getLength(), 0);
```
mronstro: is there any advantage in using std::fill instead of memset?

mronstro: I am about as positive to new things in C++ as Linus Torvalds :)

storage/ndb/rest-server/data-access-rondb/src/db-operations/pk/common.cpp
```diff
@@ -312,23 +291,24 @@ RS_Status SetOperationPKCol(const NdbDictionary::Column *col, NdbOperation *oper
   case NdbDictionary::Column::Decimal: {
```
mronstro, line 263: A generic comment is that if you are ready to return from function with RS_OK, do it immediately without adding one more if statement and one more variable. This is replicated in many case-statements

mronstro: One more generic comment, if possible ensure that the common path is that the if-condition is true. If not possible try to use unlikely.

mronstro: One more generic comment: NEVER, EVER use try-catch in C++, it is a performance bug and complicates simple code to make it simple. Was one of my first lessons in my career, try-catch can cause unsolvable issues.

storage/ndb/rest-server/data-access-rondb/src/db-operations/pk/common.cpp
```diff
@@ -364,78 +345,38 @@ RS_Status SetOperationPKCol(const NdbDictionary::Column *col, NdbOperation *oper
     if (request->PKValueNDBStr(colIdx, col, &charStr) != 0) {
       return RS_CLIENT_ERROR(ERROR_019);
     }
+
     if (operation->equal(request->PKName(colIdx), charStr, data_len) != 0) {
```
mronstro: colIdx, is this a local colIdx for the query, or is it the colIdx of the column in the table, if so, there is an equal method using colIdx directly which makes it a lot faster

storage/ndb/rest-server/data-access-rondb/src/db-operations/pk/common.cpp
```diff
@@ -364,78 +345,38 @@ RS_Status SetOperationPKCol(const NdbDictionary::Column *col, NdbOperation *oper
...
+      return RS_CLIENT_ERROR(std::string(ERROR_008) + " " +
+                             "Decoded data length is greater than column length. " +
+                             "Column: " + std::string(col->getName()) +
+                             " Length: " + std::to_string(col->getLength()));
+    }
+
+    char pk[BINARY_MAX_SIZE_IN_BYTES_DECODED];
+    std::fill(pk, pk + col->getLength(), 0);
+    std::pair<std::size_t, std::size_t> ret =
+        boost::beast::detail::base64::decode(pk, encoded_str, encoded_str_len);
```
mronstro, line 346: Not sure I am a fan of using Boost library more than necessary, pretty sure there is a Base64 function in MySQL already, otherwise I have one myself as well. Using Boost makes it impossible to optimise this function in the future.

storage/ndb/rest-server/data-access-rondb/src/db-operations/pk/common.cpp
```diff
@@ -364,78 +345,38 @@ RS_Status SetOperationPKCol(const NdbDictionary::Column *col, NdbOperation *oper
...
-            " ret.first (" + std::to_string(ret.first) +
-            ") is not equal to decoded_size (" + std::to_string(decoded_size)
-        );
+    /// Binary data is sent as base64 string
+    assert(col->getLength() <= BINARY_MAX_SIZE_IN_BYTES);
+    const char *encoded_str      = request->PKValueCStr(colIdx);
+    const size_t encoded_str_len = request->PKValueLen(colIdx);
+    size_t col_len               = col->getLength();
+
+    size_t decoded_size = boost::beast::detail::base64::decoded_size(encoded_str_len);
```
mronstro, line 334: I have an implementation of decode Base64 that we can easily convert such that it does both decode and checks that the length is ok. With this manner of doing it you will have to scan the binary array twice, one to calculate the length and one to decode.

So let's change to this version.

build_scripts/build_script_debug.sh
```diff
@@ -2,5 +2,5 @@ SRC_DIR=$1
 if [ -z $SRC_DIR ]; then
     SRC_DIR=./..
 fi
-cmake -DWITH_DEBUG=1 -DWITH_SSL=$OPENSSL_ROOT -DWITH_RDRS=1 -DWITH_NDBCLUSTER=1 -DWITH_NDB_TEST=1 -DWITH_NDBAPI_EXAMPLES=1 -DWITH_ROUTER=0 -DWITH_UNIT_TESTS=1 -DWITH_ERROR_INSERT=1 -DWITH_BOOST=$BOOST_ROOT -DWITH_NDB_JAVA=1 -DWITH_LIBEVENT=bundled $SRC_DIR
-make -j16
+cmake -DWITH_DEBUG=1 -DWITH_SSL=$OPENSSL_ROOT -DWITH_RDRS=1 -DWITH_NDBCLUSTER=1 -DWITH_NDB_TEST=1 -DWITH_NDBAPI_EXAMPLES=1 -DWITH_ROUTER=0 -DWITH_UNIT_TESTS=1 -DWITH_ERROR_INSERT=1 -DWITH_BOOST=$BOOST_ROOT -DWITH_NDB_JAVA=1 -DWITH_LIBEVENT=bundled -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE $SRC_DIR
+make -j$(nproc)
```
mronstro, line 6: Maybe a good idea to make the same change to other build scripts

storage/ndb/rest-server/data-access-rondb/src/db-operations/pk/common.cpp
```diff
@@ -631,22 +537,27 @@ RS_Status SetOperationPKCol(const NdbDictionary::Column *col, NdbOperation *oper
...
-    size_t col_byte_size = col->getSizeInBytes();
-    int precision     = col->getPrecision();
-    unsigned char packed[6];
+    size_t col_size = col->getSizeInBytes();
+    int precision   = col->getPrecision();
+    unsigned char packed[TIME2_MAX_SIZE_IN_BYTES];
+
+    int warnings = 0;
+    my_datetime_adjust_frac(&l_time, precision, &warnings, true);
+    assert(warnings == 0);
```
mronstro: what happens in production compilation when warnings != 0?

storage/ndb/rest-server/data-access-rondb/src/db-operations/pk/common.cpp
```diff
@@ -631,22 +537,27 @@ RS_Status SetOperationPKCol(const NdbDictionary::Column *col, NdbOperation *oper
...
-    if ( exitCode != 0) {
+    if (operation->equal(request->PKName(colIdx), reinterpret_cast<char *>(packed), col_size) !=
+        0) {
       return RS_SERVER_ERROR(ERROR_023);
     }
     return RS_OK;
   }
   case NdbDictionary::Column::Datetime2: {
     ///< 5 bytes plus 0-3 fraction
+    assert(col->getSizeInBytes() <= DATETIME_MAX_SIZE_IN_BYTES);
```
mronstro: what happens in production code when this assert is untrue?

storage/ndb/rest-server/data-access-rondb/src/db-operations/pk/common.cpp
```diff
@@ -658,28 +569,31 @@ RS_Status SetOperationPKCol(const NdbDictionary::Column *col, NdbOperation *oper
...
+    if (operation->equal(request->PKName(colIdx), reinterpret_cast<char *>(packed), col_size) !=
+        0) {
       return RS_SERVER_ERROR(ERROR_023);
     }
     return RS_OK;
   }
   case NdbDictionary::Column::Timestamp2: {
     // epoch range 0 , 2147483647
     /// < 4 bytes + 0-3 fraction
+    assert(col->getSizeInBytes() <= TIMESTAMP2_MAX_SIZE_IN_BYTES);
```
mronstro: What happens ....

storage/ndb/rest-server/data-access-rondb/src/db-operations/pk/common.cpp
```diff
@@ -725,17 +639,21 @@ RS_Status SetOperationPKCol(const NdbDictionary::Column *col, NdbOperation *oper
     // TODO(salman) how to deal with time zone setting in mysql server
     //
 
-    TruncatePrecision(&l_time, status.fractional_digits, precision);
+    int warnings = 0;
+    my_datetime_adjust_frac(&l_time, precision, &warnings, true);
+    assert(warnings == 0);
```
mronstro: What happens ...

storage/ndb/rest-server/data-access-rondb/src/db-operations/pk/common.cpp
```diff
@@ -888,19 +807,20 @@ RS_Status WriteColToRespBuff(const NdbRecAttr *attr, PKRResponse *response) {
   }
   case NdbDictionary::Column::Bit: {
     //< Bit, length specifies no of bits
-    Uint32 words     = attr->getColumn()->getLength() / 8;
+    Uint32 words = attr->getColumn()->getLength() / 8;
     if (attr->getColumn()->getLength() % 8 != 0) {
       words += 1;
     }
+    assert(words <= BIT_MAX_SIZE_IN_BYTES);
```
mronstro: What happens ...

mronstro (commented):
Gone through the PR, change the most important things and make a plan for how to change more long-term changes.

mronstro (approved):

PR #249, opened 2023-04-05
RONDB-332: API Key Cache Fixes

smkniazi:


storage/ndb/rest-server/data-access-rondb/src/rdrs-hopsworks-dal.cpp
```diff
@@ -533,19 +567,41 @@ RS_Status find_all_projects(int uid, char ***projects, int *count) {
...
+  do {
+    project_vec.clear();
+    status = find_all_projects_int(uid, &project_vec);
+    orc++;
+    if (status.http_code == SUCCESS || orc > OP_RETRY_COUNT || !CanRetryOperation(status)) {
+      break;
+    }
+    usleep(ExponentialDelayWithJitter(orc, OP_RETRY_INITIAL_DELAY_IN_MS, OP_RETRY_JITTER_IN_MS) *
+           1000);
+  } while (true);
```
olapiv: This is the third time I've seen this pattern - I'm sure you can extract it?

storage/ndb/rest-server/rest-api-server/README.md
```diff
@@ -207,16 +207,23 @@ Currently, the REST API server only supports [Hopsworks API Keys](https://docs.h
...
-                "UseHopsworksAPIKeys": true,
-                "HopsworksAPIKeysCacheValiditySec": 3
+                 "TLS": {
+                         "EnableTLS": true,
+                         "RequireAndVerifyClientCert": false,
+                         "CertificateFile": "",
+                         "PrivateKeyFile": "",
+                         "RootCACertFile": ""
+                 }
+                 "APIKeyParameters": {
```
olapiv: "Parameters" is kinda implicit

storage/ndb/rest-server/rest-api-server/internal/config/structs.go
```diff
@@ -25,12 +25,43 @@ import (
...
+var _ Validater = (*REST)(nil)
+var _ Validater = (*MySQLServer)(nil)
+var _ Validater = (*Testing)(nil)
+var _ Validater = (*MySQL)(nil)
+var _ Validater = (*Mgmd)(nil)
+var _ Validater = (*RonDB)(nil)
+var _ Validater = (*TestParameters)(nil)
+var _ Validater = (*APIKeyParameters)(nil)
+var _ Validater = (*Security)(nil)
+var _ Validater = (*TLS)(nil)
```
olapiv: Neat idea, but I don't think this is going to help, unless you dynamically run through every struct through a function that requires a Validator. I would just remove it.

Also, Validater -> Validator

storage/ndb/rest-server/rest-api-server/internal/config/getters_setters.go
```diff
@@ -23,8 +23,8 @@ import (
 	"io/ioutil"
 )
 
-func GetAll() AllConfigs {
-	return globalConfig
+func GetAll() *AllConfigs {
+	return &globalConfig
```
olapiv: No, if you do this, you have to handle concurrency issues writing to the struct.

storage/ndb/rest-server/rest-api-server/internal/handlers/batchpkread/handler.go
```diff
@@ -18,28 +18,34 @@ package batchpkread
...
 type Handler struct {
-	heap *heap.Heap
+	heap        *heap.Heap
+	apiKeyCache apikey.APIKeyCacher
 }
 
-func New(heap *heap.Heap) Handler {
-	return Handler{heap}
+// ensure all the interface methods are implemented
+var _ handlers.Handler = (*Handler)(nil)
```
olapiv: This is already taken care of in the `Handle` function in `internal/handlers/handler.go`.

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey.go
```diff
@@ -17,54 +17,72 @@
...
+	"hopsworks.ai/rdrs/internal/security/apikey/apikey_cache/hw_apikey_cache"
 )
 
+type APIKeyCacher interface {
+	ValidateAPIKey(apiKey *string, dbs ...*string) error
+	Cleanup() error
+	LastUpdated(apiKey *string) time.Time
+	LastUsed(apiKey *string) time.Time
+	Size() int
+}
```
olapiv, line 29: I think unless you're gonna implement a dummy right away (and there is a use-case for creating one), building an interface is a waste of space and makes things more difficult to read.

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_cache/hw_apikey_cache/hw_api_cache.go
```diff
@@ -0,0 +1,319 @@
...
+	"sync"
+	"time"
+
+	"hopsworks.ai/rdrs/internal/config"
+	"hopsworks.ai/rdrs/internal/dal"
+	"hopsworks.ai/rdrs/internal/log"
+	"hopsworks.ai/rdrs/internal/security/apikey/apikey_cache"
+)
+
+var _ apikey_cache.HWAPIKeyCache = (*HWAPIKeyCache)(nil)
```
olapiv: Use functions that require interfaces for implicit checks rather than creating this code everywhere. See `Handle()` method in `handler.go`.

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_cache/hw_apikey_cache/hw_api_cache.go
```diff
@@ -0,0 +1,319 @@
...
+
+func NewAPIKeyCache() apikey_cache.HWAPIKeyCache {
+	rand.Seed(time.Now().Unix())
```
olapiv: You're setting this globally for the instantiation of a local object?

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_cache/hw_apikey_cache/hw_api_cache.go
```diff
@@ -0,0 +1,319 @@
...
+
+	"hopsworks.ai/rdrs/internal/config"
+	"hopsworks.ai/rdrs/internal/dal"
+	"hopsworks.ai/rdrs/internal/log"
+	"hopsworks.ai/rdrs/internal/security/apikey/apikey_cache"
+)
+
+var _ apikey_cache.HWAPIKeyCache = (*HWAPIKeyCache)(nil)
+
+func NewAPIKeyCache() apikey_cache.HWAPIKeyCache {
```
olapiv: The return value just mystifies things. You can always convert to an interface, it's more risky to return to a struct though.

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_cache/hw_apikey_cache/hw_api_cache.go
```diff
@@ -0,0 +1,319 @@
...
+		}
+	}
+}
+
+func (hwc *HWAPIKeyCache) cacheEntryUpdater(apiKey *string, started *bool) {
+
+	udbs, ok := hwc.key2UserDBsCache[*apiKey] // not need for read lock here as the caller holds write lock
+	if !ok {
+		log.Errorf("Cache updater failed. Report programming error. API Key %s", *apiKey)
+		return
```
olapiv: Why not return an error and let the caller decide what to do?

storage/ndb/rest-server/rest-api-server/internal/handlers/batchpkread/handler.go
```diff
@@ -18,28 +18,34 @@ package batchpkread
...
+
+func New(heap *heap.Heap, apiKeyCache apikey.APIKeyCacher) Handler {
+	return Handler{heap, apiKeyCache}
 }
 
 // TODO: This might have to be instantiated properly
 var pkReadHandler pkread.Handler
 
-func (h Handler) Validate(request interface{}) error {
+func (h *Handler) Validate(request interface{}) error {
```
olapiv, line 44: These handlers are very light-weight, since they just contain pointers. I'd remove the complexity of making them pointers here as well (no nil checks necessary).

olapiv (changes requested):
I gave up after seeing all the interfaces. I'll continue again after you've addressed my comments, then I'll actually look into the logic fixes.

storage/ndb/rest-server/data-access-rondb/src/rdrs-hopsworks-dal.cpp
smkniazi: Nope. what is the issue here

storage/ndb/rest-server/rest-api-server/README.md
smkniazi: agree.

storage/ndb/rest-server/rest-api-server/internal/config/getters_setters.go
smkniazi: We read the conf in different places in the server and there is no need to return a copy of it. We should avoid creating objects as much as we can to reduce work for GC
The conf is only modified in the tests and the places where it is modified are single-threaded.

storage/ndb/rest-server/rest-api-server/internal/config/structs.go
smkniazi: Trying to follow go naming convention here. The interface for Validate will be Validater. 

I will not remove the above code as the purpose is to throw early compiler errors  if the interface change

storage/ndb/rest-server/rest-api-server/internal/handlers/batchpkread/handler.go
smkniazi: every additional copy of an object counts.

storage/ndb/rest-server/rest-api-server/internal/handlers/batchpkread/handler.go
smkniazi: Yes, kinda but not fully. It is better to have such checks to find unused and unimplemented methods

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey.go
smkniazi: It is needed. This interface will be used to build another API cache specific to Allstate.

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_cache/hw_apikey_cache/hw_api_cache.go
smkniazi: I do not see what is wrong with setting the seed here

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_cache/hw_apikey_cache/hw_api_cache.go
smkniazi: This is async job to update the cache. It can only log the errors

smkniazi (commented):
I have pushed some changes. 
I have not accommodated the change requests for checking if a struct implements an interface as it helps https://go.dev/doc/faq#implements_interface

storage/ndb/rest-server/rest-api-server/internal/config/getters_setters.go
olapiv: Yeah I understand the issue. What about parallel tests in the future? Or tests running after that test?

I was initially thinking of creating a local config object at the top and then passing it down, but then infinitely many methods/structs would need the config. Also wouldn't solve the server/client conflict in integration tests. We'll have the same problem with zap Logger.

If you want to return a pointer, then just remove the Getter and capitalise the variable?

storage/ndb/rest-server/rest-api-server/internal/config/structs.go
olapiv: If I remove a validation method, it probably has a reason.

If add a struct field to Config, nothing will happen. I have to actively add another line here as well. So basically the reviewer is the "compilation error", just like now.

storage/ndb/rest-server/rest-api-server/internal/handlers/batchpkread/handler.go
olapiv: Ok

storage/ndb/rest-server/rest-api-server/internal/handlers/batchpkread/handler.go
olapiv: Why not fully? Try to remove the Validate() method and see what happens.

storage/ndb/rest-server/data-access-rondb/src/rdrs-hopsworks-dal.cpp
olapiv: Replication. It's fine if it's too complicated.

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey.go
olapiv: Then add a comment that we need this to satisfy different API key systems. Wouldn't be happy to see AllState-specific code in the RonDB repository though. Rather a general abstraction of API key systems.

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_cache/hw_apikey_cache/hw_api_cache.go
olapiv: Then at least say why.

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_cache/hw_apikey_cache/hw_api_cache.go
olapiv: But then let the worker decide that. You don't even have to do 'err :=' if you don't want to.

storage/ndb/rest-server/rest-api-server/internal/config/getters_setters.go
olapiv: Actually, I would recommend this:
- return a value
- get the value once at server instantiation and pass its pointer to all handlers

If you really want to change the config in tests, you have to restart the server.

storage/ndb/rest-server/data-access-rondb/src/rdrs-hopsworks-dal.cpp
smkniazi: I can get rid of it using MACROS but not in this PR

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_cache/hw_apikey_cache/hw_api_cache.go
smkniazi: to randomize ...

olapiv: To randomise what? Why is it necessary?

storage/ndb/rest-server/rest-api-server/internal/config/structs.go
```diff
@@ -318,6 +299,10 @@ type AllConfigs struct {
 	Testing  Testing
 }
 
+func (c *AllConfigs) xalidate() error {
+	return nil
+}
```
olapiv: Remove this

storage/ndb/rest-server/data-access-rondb/src/rdrs-hopsworks-dal.cpp
smkniazi: fixed it

storage/ndb/rest-server/rest-api-server/internal/config/structs.go
smkniazi: removed it

storage/ndb/rest-server/rest-api-server/internal/config/getters_setters.go
smkniazi: reverted this change

storage/ndb/rest-server/rest-api-server/internal/handlers/batchpkread/handler.go
smkniazi: reverted this change

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_cache/hw_apikey_cache/hw_api_cache.go
smkniazi: reverted this change

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_cache/hw_apikey_cache/hw_api_cache.go
smkniazi: moved it to unit test, which is the correct place

smkniazi (commented):
pushed all the changes

storage/ndb/rest-server/rest-api-server/internal/dal/heap/heap.go
```diff
@@ -93,6 +94,11 @@ func (heap *Heap) releaseAllBuffers() {
 	heap.mutex.Lock()
 	defer heap.mutex.Unlock()
 
+	if heap.buffersStats.AllocationsCount != int64(len(heap.buffers)) {
+		log.Errorf("Shutting down heap. Number of free buffers do not match. Expecting: %d, Got: %d.",
```
olapiv: You're not "shutting down" anything afterwards?

storage/ndb/rest-server/rest-api-server/internal/integrationtests/initialiase_tests.go
```diff
@@ -54,14 +52,32 @@ func InitialiseTesting(conf config.AllConfigs, createOnlyTheseDBs ...string) (cl
...
+
+	//---------------------------- TLS ----------------------------------------
+	if conf.Security.TLS.EnableTLS {
+		cleanupTLSCerts, err := testutils.CreateAllTLSCerts()
 		if err != nil {
-			return cleanup, err
+			return nil, err
 		}
+		cleanupFNs[index] = &cleanupTLSCerts
+		index++
```
olapiv: Instead of using `cleanupFNs := make([]*func(), 10)`, use `cleanupFNs := []func(){}` and then
`cleanupFNs = append(cleanupFNs, cleanupX)`, then you don't have to keep track of the index. You can curry the function, such as:

```go
cleanupWrapper := func(cleanupFNs func()) {
        return func () {
            // clean up in reverse order
            for i := len(cleanupFNs) - 1; i >= 0; i-- {
                if cleanupFNs[i] != nil {
                    (cleanupFNs[i])()
                }
            }
        }
 }
```

then cleanupFNs doesn't have to be defined from the start.

storage/ndb/rest-server/rest-api-server/internal/integrationtests/initialiase_tests.go
```diff
@@ -54,14 +52,32 @@ func InitialiseTesting(conf config.AllConfigs, createOnlyTheseDBs ...string) (cl
...
+	index := 0
+	cleanupFNs := make([]*func(), 10)
+	cleanupFN := func() {
+		//clean up in reverse order
+		for i := len(cleanupFNs) - 1; i >= 0; i-- {
+			if cleanupFNs[i] != nil {
+				(*cleanupFNs[i])()
+			}
+		}
+	}
```
olapiv, line 65: I think we use this pattern in other parts of the code as well - you could extract it, it looks a bit hacky here.

storage/ndb/rest-server/rest-api-server/internal/integrationtests/initialiase_tests.go
```diff
@@ -74,61 +90,66 @@ func InitialiseTesting(conf config.AllConfigs, createOnlyTheseDBs ...string) (cl
...
+	//---------------------------- API KEY Cache ------------------------------
+	apiKeyCache, err := apikey.NewAPIKeyCache()
+	if err != nil {
+		cleanupFN()
+		return nil, fmt.Errorf("failed creating new API Key Cache; error: %v ", err)
+	}
+	apiKeyCleanup := func() {
+		apiKeyCache.Cleanup()
+	}
+	cleanupFNs[index] = &apiKeyCleanup
```
olapiv: Just reference `apiKeyCache.Cleanup`

storage/ndb/rest-server/rest-api-server/internal/integrationtests/initialiase_tests.go
```diff
@@ -74,61 +90,66 @@ func InitialiseTesting(conf config.AllConfigs, createOnlyTheseDBs ...string) (cl
...
-			cleanupServers()
-			releaseBuffers()
-			cleanupTLSCerts()
-			return cleanup, fmt.Errorf("could not create profile.out; error: %w ", err)
+			cleanupFN()
+			return nil, fmt.Errorf("could not create profile.out; error: %w ", err)
 		}
-		defer f.Close()
+		fileCloser := func() { f.Close() }
+		cleanupFNs[index] = &fileCloser
```
olapiv: Again, just reference `f.Close`.

storage/ndb/rest-server/rest-api-server/internal/integrationtests/initialiase_tests.go
```diff
@@ -74,61 +90,66 @@ func InitialiseTesting(conf config.AllConfigs, createOnlyTheseDBs ...string) (cl
...
-		// Running defer here in case checking the heap fails
-		defer cleanupTLSCerts()
-		defer releaseBuffers()
-		defer cleanupServers()
-		defer pprof.StopCPUProfile()
-
-		stats := newHeap.GetNativeBuffersStats()
-		if stats.BuffersCount != stats.FreeBuffers {
-			log.Errorf("Number of free buffers do not match. Expecting: %d, Got: %d",
-				stats.BuffersCount, stats.FreeBuffers)
```
olapiv, line 131: You don't want to include this in the cleanup for a check?

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey.go
```diff
@@ -17,54 +17,72 @@
...
+
+type APIKeyCache struct {
+	HWAPIKeyCache apikey_cache.HWAPIKeyCache
+}
+
+var _ APIKeyCacher = (*APIKeyCache)(nil)
```
olapiv: This won't be necessary, since you use the interface for the Handlers.

storage/ndb/rest-server/rest-api-server/internal/security/apikey/apikey_cache/apikey_cache.go
```diff
@@ -0,0 +1,29 @@
...
+import "time"
+
+type HWAPIKeyCache interface {
+	UpdateCache(apiKey *string) error
+	FindAndValidate(apiKey *string, dbs ...*string) (keyFoundInCache, allowedAccess bool)
+	Cleanup() error
+	LastUpdated(apiKey *string) time.Time
+	LastUsed(apiKey *string) time.Time
+	Size() int
+}
```
olapiv: Why is this an interface?

olapiv (changes requested):

PR #290, opened 2023-05-10
Add unit test to feature store API

kennethmhc:


smkniazi (approved):
Unfortunately, I could not go through it in detail. It looks OK to me. We can go through the entire code together once the first version is complete

PR #299, opened 2023-05-18
RONDB-386: Added implementation & bench results of Easyjson

olapiv:


mronstro (commented):
There is no reasoning of what the benefits of using easyjson would be.
I would like to see links to public comparisons with other libraries. What benefits
does it have compared to simdjson and so forth.

olapiv:
Well the main reason is heavily reduced amount of memory allocations (-25% for few columns). I haven't done any profiling yet, but afaik it reduces the amount of reflection that is done with the standard library.

I got wind of this library through this blog: https://yalantis.com/blog/speed-up-json-encoding-decoding/
That does not mention simdjson though.

I'm not saying this should be merged, but perhaps benchmarking it with VMs could help us determine any (dis-)advantages IRL.

PR #314, opened 2023-06-05
[FSTORE-709] Feature store API

kennethmhc:


MYSQL_VERSION
```diff
@@ -1,4 +1,4 @@
 MYSQL_VERSION_MAJOR=21
 MYSQL_VERSION_MINOR=04
-MYSQL_VERSION_PATCH=14
+MYSQL_VERSION_PATCH=99
```
smkniazi: 99 --> 14

storage/ndb/rest-server/data-access-rondb/src/feature_store/feature_store.cpp
```diff
@@ -0,0 +1,910 @@
+/*
```
smkniazi, line 1:
```c++
diff --git a/storage/ndb/rest-server/data-access-rondb/src/feature_store/feature_store.cpp b/storage/ndb/rest-server/data-access-rondb/src/feature_store/feature_store.cpp
index 316bdbda3b28..53e715be935a 100644
--- a/storage/ndb/rest-server/data-access-rondb/src/feature_store/feature_store.cpp
+++ b/storage/ndb/rest-server/data-access-rondb/src/feature_store/feature_store.cpp
@@ -71,13 +71,13 @@ RS_Status find_project_id_int(Ndb *ndb_object, const char *feature_store_name, I
   int projectname_col_id      = table_dict->getColumn("projectname")->getColumnNo();
   Uint32 projectname_col_size = (Uint32)table_dict->getColumn("projectname")->getSizeInBytes();
   assert(projectname_col_size == PROJECT_PROJECTNAME_SIZE);
-  if (strlen(feature_store_name) >=
-      PROJECT_PROJECTNAME_SIZE - 1) {  // col_size include length byte(s)
+  int feature_store_name_len = strlen(feature_store_name);
+  if (feature_store_name_len >= PROJECT_PROJECTNAME_SIZE - 1) {  // col_size include length byte(s)
     return RS_CLIENT_ERROR("Wrong length of the projectname");
   }
 
   char cmp_str[PROJECT_PROJECTNAME_SIZE];
-  memcpy(cmp_str + 1, feature_store_name, PROJECT_PROJECTNAME_SIZE - 1);
+  memcpy(cmp_str + 1, feature_store_name, feature_store_name_len + 1);  //+1 for null
   cmp_str[0] = static_cast<char>(strlen(feature_store_name));
 
   NdbScanFilter filter(scan_op);
@@ -185,13 +185,13 @@ RS_Status find_feature_store_id_int(Ndb *ndb_object, const char *feature_store_n
   int name_col_id      = table_dict->getColumn("name")->getColumnNo();
   Uint32 name_col_size = (Uint32)table_dict->getColumn("name")->getSizeInBytes();
   assert(name_col_size == FEATURE_STORE_NAME_SIZE);
-  if (strlen(feature_store_name) >=
-      FEATURE_STORE_NAME_SIZE - 1) {  // col_size include length byte(s)
+  int feature_store_name_len = strlen(feature_store_name);
+  if (feature_store_name_len >= FEATURE_STORE_NAME_SIZE - 1) {  // col_size include length byte(s)
     return RS_CLIENT_ERROR("Wrong length of column name");
   }
 
   char cmp_str[FEATURE_STORE_NAME_SIZE];
-  memcpy(cmp_str + 1, feature_store_name, FEATURE_STORE_NAME_SIZE - 1);
+  memcpy(cmp_str + 1, feature_store_name, feature_store_name_len + 1);  // +1 for null
   cmp_str[0] = static_cast<char>(strlen(feature_store_name));
 
   NdbScanFilter filter(scan_op);
@@ -303,12 +303,13 @@ RS_Status find_feature_view_id_int(Ndb *ndb_object, int feature_store_id,
   int name_col_id      = table_dict->getColumn("name")->getColumnNo();
   Uint32 name_col_size = (Uint32)table_dict->getColumn("name")->getSizeInBytes();
   assert(name_col_size == FEATURE_VIEW_NAME_SIZE);
-  if (strlen(feature_view_name) >= FEATURE_VIEW_NAME_SIZE - 1) {  // col_size include length byte(s)
+  int feature_view_name_len = strlen(feature_view_name);
+  if (feature_view_name_len >= FEATURE_VIEW_NAME_SIZE - 1) {  // col_size include length byte(s)
     return RS_CLIENT_ERROR("Wrong length of column name");
   }
 
   char cmp_str[FEATURE_VIEW_NAME_SIZE];
-  memcpy(cmp_str + 1, feature_view_name, FEATURE_VIEW_NAME_SIZE - 1);
+  memcpy(cmp_str + 1, feature_view_name, feature_view_name_len + 1);  // +1 for null
   cmp_str[0] = static_cast<char>(strlen(feature_view_name));
 
   NdbScanFilter filter(scan_op);
@@ -412,7 +413,8 @@ RS_Status find_feature_view_id(int feature_store_id, const char *feature_view_na
 
 //-------------------------------------------------------------------------------------------------
 
-RS_Status find_training_dataset_join_data_int(Ndb *ndb_object, int feature_view_id, Training_Dataset_Join **tdjs, int *tdjs_size) {
+RS_Status find_training_dataset_join_data_int(Ndb *ndb_object, int feature_view_id,
+                                              Training_Dataset_Join **tdjs, int *tdjs_size) {
   NdbError ndb_err;
   const NdbDictionary::Table *table_dict;
   NdbTransaction *tx;
@@ -453,15 +455,15 @@ RS_Status find_training_dataset_join_data_int(Ndb *ndb_object, int feature_view_
     return RS_RONDB_SERVER_ERROR(ndb_err, ERROR_031);
   }
 
-  NdbRecAttr *td_join_id_attr       = scan_op->getValue("id", nullptr);
-  NdbRecAttr *prefix_attr           = scan_op->getValue("prefix", nullptr);
+  NdbRecAttr *td_join_id_attr = scan_op->getValue("id", nullptr);
+  NdbRecAttr *prefix_attr     = scan_op->getValue("prefix", nullptr);
 
   if (td_join_id_attr == nullptr || prefix_attr == nullptr) {
     return RS_RONDB_SERVER_ERROR(ndb_err, ERROR_019);
   }
 
   assert(TRAINING_DATASET_JOIN_PREFIX_SIZE ==
-        (Uint32)table_dict->getColumn("prefix")->getSizeInBytes());
+         (Uint32)table_dict->getColumn("prefix")->getSizeInBytes());
 
   if (tx->execute(NdbTransaction::NoCommit) != 0) {
     ndb_err = ndb_object->getNdbError();
@@ -469,29 +471,29 @@ RS_Status find_training_dataset_join_data_int(Ndb *ndb_object, int feature_view_
     return RS_RONDB_SERVER_ERROR(ndb_err, ERROR_009);
   }
 
-  bool check   = 0;
+  bool check = 0;
   std::vector<Training_Dataset_Join> tdjsv;
-while ((check = scan_op->nextResult(true)) == 0) {
-  do {
-    Training_Dataset_Join tdj;
-    tdj.id = td_join_id_attr->int32_value();
-    
-    if (prefix_attr->isNULL()) {
-      tdj.prefix[0] = 0;
-    } else {
-      Uint32 prefix_attr_bytes;
-      const char *prefix_data_start = nullptr;
-      if (GetByteArray(prefix_attr, &prefix_data_start, &prefix_attr_bytes) != 0) {
-        return RS_CLIENT_ERROR(ERROR_019);
+  while ((check = scan_op->nextResult(true)) == 0) {
+    do {
+      Training_Dataset_Join tdj;
+      tdj.id = td_join_id_attr->int32_value();
+
+      if (prefix_attr->isNULL()) {
+        tdj.prefix[0] = 0;
+      } else {
+        Uint32 prefix_attr_bytes;
+        const char *prefix_data_start = nullptr;
+        if (GetByteArray(prefix_attr, &prefix_data_start, &prefix_attr_bytes) != 0) {
+          return RS_CLIENT_ERROR(ERROR_019);
+        }
+
+        memcpy(tdj.prefix, prefix_data_start, prefix_attr_bytes);
+        tdj.prefix[prefix_attr_bytes] = 0;
       }
 
-      memcpy(tdj.prefix, prefix_data_start, prefix_attr_bytes);
-      tdj.prefix[prefix_attr_bytes] = 0;
-    }
-    
-    tdjsv.push_back(tdj);
-  } while ((check = scan_op->nextResult(false)) == 0);
-} 
+      tdjsv.push_back(tdj);
+    } while ((check = scan_op->nextResult(false)) == 0);
+  }
 
   // check for errors happened during the reading process
   NdbError error = scan_op->getNdbError();
@@ -508,13 +510,13 @@ while ((check = scan_op->nextResult(true)) == 0) {
     return RS_CLIENT_404_ERROR();
   }
 
-    // freed by CGO
+  // freed by CGO
   *tdjs_size = tdjsv.size();
   void *ptr  = (Training_Dataset_Join *)malloc(tdjsv.size() * sizeof(Training_Dataset_Join));
   *tdjs      = (Training_Dataset_Join *)ptr;
   for (Uint64 i = 0; i < tdjsv.size(); i++) {
-    (*tdjs + i)->id = tdjsv[i].id; 
-    memcpy((*tdjs + i)->prefix , tdjsv[i].prefix, strlen(tdjsv[i].prefix)+1);
+    (*tdjs + i)->id = tdjsv[i].id;
+    memcpy((*tdjs + i)->prefix, tdjsv[i].prefix, strlen(tdjsv[i].prefix) + 1);
   }
   tdjsv.clear();
 
@@ -528,7 +530,8 @@ while ((check = scan_op->nextResult(true)) == 0) {
  * SELECT id AS td_join_id, feature_group AS feature_group_id, prefix FROM training_dataset_join
  * WHERE feature_view_id = {feature_view_id}
  */
-RS_Status find_training_dataset_join_data(int feature_view_id, Training_Dataset_Join **tdjs, int *tdjs_size) {
+RS_Status find_training_dataset_join_data(int feature_view_id, Training_Dataset_Join **tdjs,
+                                          int *tdjs_size) {
 
   Ndb *ndb_object  = nullptr;
   RS_Status status = rdrsRonDBConnection->GetNdbObject(&ndb_object);
@@ -549,7 +552,8 @@ RS_Status find_training_dataset_join_data(int feature_view_id, Training_Dataset_
 //-------------------------------------------------------------------------------------------------
 
 RS_Status find_feature_group_data_int(Ndb *ndb_object, int feature_group_id, char *name,
-                                      int *online_enabled, int *feature_store_id, int *feature_group_version) {
+                                      int *online_enabled, int *feature_store_id,
+                                      int *feature_group_version) {
 
   NdbError ndb_error;
   const NdbDictionary::Table *table_dict;
@@ -582,13 +586,14 @@ RS_Status find_feature_group_data_int(Ndb *ndb_object, int feature_group_id, cha
     return RS_SERVER_ERROR(ERROR_023);
   }
 
-  NdbRecAttr *name_attr           = ndb_op->getValue("name", nullptr);
-  NdbRecAttr *online_enabled_attr  = ndb_op->getValue("online_enabled", nullptr);
-  NdbRecAttr *feature_store_id_attr = ndb_op->getValue("feature_store_id", nullptr);
+  NdbRecAttr *name_attr                  = ndb_op->getValue("name", nullptr);
+  NdbRecAttr *online_enabled_attr        = ndb_op->getValue("online_enabled", nullptr);
+  NdbRecAttr *feature_store_id_attr      = ndb_op->getValue("feature_store_id", nullptr);
   NdbRecAttr *feature_group_version_attr = ndb_op->getValue("version", nullptr);
   assert(FEATURE_GROUP_NAME_SIZE == (Uint32)table_dict->getColumn("name")->getSizeInBytes());
 
-  if (name_attr == nullptr || online_enabled_attr == nullptr || feature_store_id_attr == nullptr || feature_group_version_attr == nullptr) {
+  if (name_attr == nullptr || online_enabled_attr == nullptr || feature_store_id_attr == nullptr ||
+      feature_group_version_attr == nullptr) {
     return RS_RONDB_SERVER_ERROR(ndb_error, ERROR_019);
   }
 
@@ -603,8 +608,8 @@ RS_Status find_feature_group_data_int(Ndb *ndb_object, int feature_group_id, cha
     return RS_CLIENT_404_ERROR();
   }
 
-  *online_enabled   = online_enabled_attr->u_8_value();
-  *feature_store_id = feature_store_id_attr->int32_value();
+  *online_enabled        = online_enabled_attr->u_8_value();
+  *feature_store_id      = feature_store_id_attr->int32_value();
   *feature_group_version = feature_group_version_attr->int32_value();
 
   Uint32 name_attr_bytes;
@@ -777,16 +782,16 @@ RS_Status find_training_dataset_data_int(Ndb *ndb_object, int feature_view_id,
   void *ptr  = (Training_Dataset_Feature *)malloc(tdfsv.size() * sizeof(Training_Dataset_Feature));
   *tdfs      = (Training_Dataset_Feature *)ptr;
   for (Uint64 i = 0; i < tdfsv.size(); i++) {
-    (*tdfs + i)->feature_id = tdfsv[i].feature_id; 
-    (*tdfs + i)->training_dataset = tdfsv[i].training_dataset; 
-    (*tdfs + i)->feature_group_id = tdfsv[i].feature_group_id; 
-    (*tdfs + i)->td_join_id = tdfsv[i].td_join_id; 
-    (*tdfs + i)->idx = tdfsv[i].idx; 
-    (*tdfs + i)->label = tdfsv[i].label; 
-    (*tdfs + i)->transformation_function_id = tdfsv[i].transformation_function_id; 
-    (*tdfs + i)->feature_view_id = tdfsv[i].feature_view_id; 
-    memcpy((*tdfs + i)->name , tdfsv[i].name, strlen(tdfsv[i].name)+1);
-    memcpy((*tdfs + i)->data_type , tdfsv[i].data_type, strlen(tdfsv[i].data_type)+1);
+    (*tdfs + i)->feature_id                 = tdfsv[i].feature_id;
+    (*tdfs + i)->training_dataset           = tdfsv[i].training_dataset;
+    (*tdfs + i)->feature_group_id           = tdfsv[i].feature_group_id;
+    (*tdfs + i)->td_join_id                 = tdfsv[i].td_join_id;
+    (*tdfs + i)->idx                        = tdfsv[i].idx;
+    (*tdfs + i)->label                      = tdfsv[i].label;
+    (*tdfs + i)->transformation_function_id = tdfsv[i].transformation_function_id;
+    (*tdfs + i)->feature_view_id            = tdfsv[i].feature_view_id;
+    memcpy((*tdfs + i)->name, tdfsv[i].name, strlen(tdfsv[i].name) + 1);
+    memcpy((*tdfs + i)->data_type, tdfsv[i].data_type, strlen(tdfsv[i].data_type) + 1);
   }
   tdfsv.clear();
 
@@ -851,7 +856,7 @@ RS_Status find_feature_store_data_int(Ndb *ndb_object, int feature_store_id, cha
     return RS_SERVER_ERROR(ERROR_023);
   }
 
-  NdbRecAttr *name_attr           = ndb_op->getValue("name", nullptr);
+  NdbRecAttr *name_attr = ndb_op->getValue("name", nullptr);
   assert(FEATURE_STORE_NAME_SIZE == (Uint32)table_dict->getColumn("name")->getSizeInBytes());
 
   if (name_attr == nullptr) {

```

smkniazi, line 1: https://repo.hops.works/dev/salman/feature_store.cpp.patch

storage/ndb/rest-server/rest-api-server/Makefile
```diff
@@ -42,7 +42,7 @@ COVERAGE_HTML_OUTPUT_FILE?=./coverage.html
 COVERAGE_FUNC_OUTPUT_FILE?=/dev/stdout
 
 #CGO_LDFLAGS
-MY_CGO_LDFLAGS?="-g -O3 -L${RDRS_LIB_DIR}/ -Wl,-rpath,../lib -Wl,-rpath,${RDRS_LIB_DIR}  -lrdrclient"
+MY_CGO_LDFLAGS?="-g -O3 -L${RDRS_LIB_DIR}/ -Wl,-rpath ${RDRS_LIB_DIR} -lrdrclient"
```
smkniazi:
```suggestion
MY_CGO_LDFLAGS?="-g -O3 -L${RDRS_LIB_DIR}/ -Wl,-rpath,../lib -Wl,-rpath,${RDRS_LIB_DIR} -lrdrclient
```

smkniazi: take the changes from upstream

storage/ndb/rest-server/rest-api-server/internal/feature_store/featurestoreerrocode.go
```diff
@@ -0,0 +1,92 @@
...
+
+type RestErrorCode struct {
+	code    int
+	reason  string
+	status  int
+	message string
+}
+
+func (err *RestErrorCode) NewMessage(msg string) *RestErrorCode {
+	var errCopy = *err
```
smkniazi, line 33: copying pointers does not create a copy of the object. Initialize a new variable for RestErrorCode

storage/ndb/rest-server/rest-api-server/internal/feature_store/featurestoreerrocode.go
```diff
@@ -0,0 +1,92 @@
...
+	var m = map[string]interface{}{
+		"code":    err.code,
+		"reason":  err.reason,
+		"message": err.message,
+	}
+	strBytes, jsonErr := json.MarshalIndent(m, "", "\t")
+	if jsonErr != nil {
+		return fmt.Sprintf("Failed to marshal FeatureStoreErrorResponse. Error: %s", jsonErr)
+	} else {
+		return fmt.Sprintf("%s", strBytes)
```
smkniazi:
```suggestion
		return string(strBytes)
```

storage/ndb/rest-server/rest-api-server/internal/feature_store/metadata.go
```diff
@@ -0,0 +1,272 @@
...
+		}
+		return nil, (*FS_NOT_EXIST).NewMessage(err.Error())
+	}
+
+	fvID, err := dal.GetFeatureViewID(fsID, featureViewName, featureViewVersion)
+	if err != nil {
+		if strings.Contains(err.Error(), "Not Found") {
+			return nil, FV_NOT_EXIST
+		}
+		return nil, (*&FV_READ_FAIL).NewMessage(err.VerboseError())
```
smkniazi:
```suggestion
		return nil, FV_READ_FAIL.NewMessage(err.VerboseError())
```

storage/ndb/rest-server/rest-api-server/internal/feature_store/metadata.go
```diff
@@ -0,0 +1,272 @@
...
+	for i, tdf := range tdfs {
+		featureGroupName, _, fsId, fgVersion, err := dal.GetFeatureGroupData(tdf.FeatureGroupID)
+		if err != nil {
+			if strings.Contains(err.Error(), "Not Found") {
+				return nil, FG_NOT_EXIST
+			}
+			return nil, (*FG_READ_FAIL).NewMessage(err.VerboseError())
+		}
+		feature := FeatureMetadata{}
+		if featureStoreName, exist := fsIdToName[fsId]; exist {
```
smkniazi:
```suggestion
		if _, exist := fsIdToName[fsId]; exist {
```

storage/ndb/rest-server/rest-api-server/internal/feature_store/metadata.go
```diff
@@ -0,0 +1,272 @@
...
+	if err != nil {
+		return nil, (*TD_JOIN_READ_FAIL).NewMessage(err.VerboseError())
+	}
+	for _, tdj := range tdJoins {
+		joinIdToPrefix[tdj.Id] = tdj.Prefix
+	}
+
+	tdfs, err := dal.GetTrainingDatasetFeature(fvID)
+	if err != nil {
+		return nil, (*TD_FEATURE_READ_FAIL).NewMessage(err.VerboseError())
```
smkniazi:
```suggestion
		return nil, TD_FEATURE_READ_FAIL.NewMessage(err.VerboseError())
```

storage/ndb/rest-server/rest-api-server/internal/feature_store/metadata.go
```diff
@@ -0,0 +1,272 @@
...
+		if strings.Contains(err.Error(), "Not Found") {
+			return nil, FV_NOT_EXIST
+		}
+		return nil, (*&FV_READ_FAIL).NewMessage(err.VerboseError())
+	}
+
+	joinIdToPrefix := make(map[int]string)
+	tdJoins, err := dal.GetTrainingDatasetJoinData(fvID)
+	if err != nil {
+		return nil, (*TD_JOIN_READ_FAIL).NewMessage(err.VerboseError())
```
smkniazi:
```suggestion
		return nil, TD_JOIN_READ_FAIL.NewMessage(err.VerboseError())
```

storage/ndb/rest-server/rest-api-server/internal/feature_store/metadata.go
```diff
@@ -0,0 +1,272 @@
...
+	features := make([]*FeatureMetadata, len(tdfs))
+	fsIdToName := make(map[int]string)
+
+	for i, tdf := range tdfs {
+		featureGroupName, _, fsId, fgVersion, err := dal.GetFeatureGroupData(tdf.FeatureGroupID)
+		if err != nil {
+			if strings.Contains(err.Error(), "Not Found") {
+				return nil, FG_NOT_EXIST
+			}
+			return nil, (*FG_READ_FAIL).NewMessage(err.VerboseError())
```
smkniazi:
```suggestion
			return nil, FG_READ_FAIL.NewMessage(err.VerboseError())
```

storage/ndb/rest-server/rest-api-server/internal/feature_store/metadata.go
```diff
@@ -0,0 +1,272 @@
...
+		feature := FeatureMetadata{}
+		if featureStoreName, exist := fsIdToName[fsId]; exist {
+			feature.FeatureStoreName = fsIdToName[fsId]
+		} else {
+			featureStoreName, err = dal.GetFeatureStoreName(fsId)
+			if err != nil {
+				if strings.Contains(err.Error(), "Not Found") {
+					return nil, FS_NOT_EXIST
+				}
+				return nil, (*FS_READ_FAIL).NewMessage(err.VerboseError())
```
smkniazi:
```suggestion
				return nil, FS_READ_FAIL.NewMessage(err.VerboseError())
```

storage/ndb/rest-server/rest-api-server/internal/handlers/batchfeaturestore/handler.go
```diff
@@ -0,0 +1,204 @@
...
+	var featureStatus = make([]api.FeatureStatus, len(*fsReq.Entries))
+	var numPassed = checkStatus(fsReq, metadata, &featureStatus)
+	var readParams = getBatchPkReadParamsMutipleEntries(metadata, fsReq.Entries, &featureStatus)
+	fsResp := response.(*api.BatchFeatureStoreResponse)
+	var features *[][]interface{}
+	if len(*readParams) > 0 {
+		ronDbErr := h.dbBatchReader.Validate(readParams)
+		if ronDbErr != nil {
+			if log.IsDebug() {
+				log.Debugf("Rondb validation failed: %s", ronDbErr.Error())
```
smkniazi:
```suggestion
				log.Debugf("RonDB validation failed: %s", ronDbErr.Error())
```

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+			featureMetadata.Type = &ft
+		}
+		featureMetadataArray[metadata.Index] = &featureMetadata
+	}
+	return &featureMetadataArray
+}
+
+func TranslateRonDbError(code int, err string) *feature_store.RestErrorCode {
+	var fsError *feature_store.RestErrorCode
+	if strings.Contains(err, "Wrong data type.") {
```
smkniazi: What if the error string change in the c code?
look at storage/ndb/rest-server/rest-api-server/internal/common/cerrors.go for reading constants defined on C side

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+		regex := regexp.MustCompile(`Expecting (\w+)\. Column: (\w+)`)
```
smkniazi, line 262: Can you avoid using regexp. Regexp are very heavy wight and it could be performance bottleneck

storage/ndb/rest-server/rest-api-server/internal/handlers/batchfeaturestore/handler.go
```diff
@@ -0,0 +1,204 @@
...
+	_, err := h.fvMetaCache.Get(
+		*fsReq.FeatureStoreName, *fsReq.FeatureViewName, *fsReq.FeatureViewVersion)
+	if err != nil {
+		return err.GetError()
+	}
+	if len(*fsReq.Entries) == 0 {
+		return feature_store.NO_PRIMARY_KEY_GIVEN.GetError()
+	}
+	if len(*fsReq.PassedFeatures) != 0 && len(*fsReq.Entries) != len(*fsReq.PassedFeatures) {
+		return feature_store.INCORRECT_PASSED_FEATURE.NewMessage("Length of passed feature does not equal to that of entries.").GetError()
```
smkniazi: "does not equal to that of entries."  what entries?

storage/ndb/rest-server/rest-api-server/internal/handlers/batchfeaturestore/handler.go
```diff
@@ -0,0 +1,204 @@
...
+			if log.IsDebug() {
+				log.Debugf("Rondb validation failed: %s", ronDbErr.Error())
+			}
+			var fsError = fshandler.TranslateRonDbError(http.StatusBadRequest, ronDbErr.Error())
+			return fsError.GetStatus(), fsError.GetError()
+		}
+		var dbResponseIntf = getPkReadResponseJSON(numPassed, *metadata)
+		code, ronDbErr := h.dbBatchReader.Execute(readParams, *dbResponseIntf)
+		if log.IsDebug() {
+			log.Debugf("Rondb response: code: %d, error: %s, body: %s", code, ronDbErr, (*dbResponseIntf).String())
```
smkniazi:
```suggestion
			log.Debugf("RonDB response: code: %d, error: %s, body: %s", code, ronDbErr, (*dbResponseIntf).String())
```

storage/ndb/rest-server/rest-api-server/internal/handlers/batchfeaturestore/handler.go
```diff
@@ -0,0 +1,204 @@
...
+
+func getFeatureValuesMultipleEntries(batchResponse *api.BatchOpResponse, entries *[]*map[string]*json.RawMessage, featureView *feature_store.FeatureViewMetadata, batchStatus *[]api.FeatureStatus) *[][]interface{} {
+	jsonResponse := (*batchResponse).String()
+	rondbResp := api.BatchResponseJSON{}
+	json.Unmarshal([]byte(jsonResponse), &rondbResp)
+	ronDbBatchResult := make([][]*api.PKReadResponseWithCodeJSON, len(*batchStatus))
+	batchResult := make([][]interface{}, len(*batchStatus))
+	for _, response := range *rondbResp.Result {
+		splitOperationId := strings.Split(*response.Body.OperationID, SEQUENCE_SEPARATOR)
+		seqNum, _ := strconv.Atoi(splitOperationId[0])
```
smkniazi: do not swallow error. handle the the error

storage/ndb/rest-server/rest-api-server/internal/handlers/batchfeaturestore/handler.go
```diff
@@ -0,0 +1,204 @@
...
+func getPkReadResponseJSON(numEntries int, metadata feature_store.FeatureViewMetadata) *api.BatchOpResponse {
+	response := (api.BatchOpResponse)(&api.BatchResponseJSON{})
+	response.Init(len(metadata.FeatureGroupFeatures) * numEntries)
+	return &response
+}
+
+func fillPassedFeaturesMultipleEntries(features *[][]interface{}, passedFeatures *[]*map[string]*json.RawMessage, featureMetadata *map[string]*feature_store.FeatureMetadata, indexLookup *map[string]int, status *[]api.FeatureStatus) {
+	if len(*passedFeatures) != 0 {
+		for i, feature := range *features {
+			// TODO: validate length of pass features
```
smkniazi: TODO ?

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+	"hopsworks.ai/rdrs/internal/security/apikey"
+	"hopsworks.ai/rdrs/pkg/api"
+)
+
+const (
+	JSON_NUMBER  = "NUMBER"
+	JSON_STRING  = "STRING"
+	JSON_BOOLEAN = "BOOLEAN"
+	JSON_NIL     = "NIL"
+	JSON_OTHER   = "OTHER"
```
smkniazi, line 41: I thought "OTHER" meant other datatypes such as arrays and objects. However, from the code, it seems it is more appropriate to call the UNSUPPORTED_TYPE

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+		if index, ok := (featureView.FeatureIndexLookup)[indexKey]; ok {
+			featureValues[index] = value
+		}
+	}
+	return &featureValues, status
+}
+
+func GetBatchPkReadParams(metadata *feature_store.FeatureViewMetadata, entries *map[string]*json.RawMessage) *[]*api.PKReadParams {
+
+	var batchReadParams = make([]*api.PKReadParams, 0, 0)
```
smkniazi:
```suggestion
	var batchReadParams = make([]*api.PKReadParams, 0)
```

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+	for _, fgFeature := range metadata.FeatureGroupFeatures {
+		testDb := fgFeature.FeatureStoreName
+		testTable := fmt.Sprintf("%s_%d", fgFeature.FeatureGroupName, fgFeature.FeatureGroupVersion)
+		var filters = make([]api.Filter, 0, 0)
```
smkniazi:
```suggestion
		var filters = make([]api.Filter, 0)
```

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+		var columns = make([]api.ReadColumn, 0, 0)
```
smkniazi:
```suggestion
		var columns = make([]api.ReadColumn, 0)
```

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+		for _, feature := range fgFeature.Features {
+			if value, ok := (*entries)[feature.Prefix+feature.Name]; ok {
+				var filter = api.Filter{Column: &feature.Name, Value: value}
+				filters = append(filters, filter)
```
smkniazi: This is critical path and "append"  is a serious performance bottleneck. 
make an array of fixed size.  for example in line 318  make([]api.Filter, len(fgFeature.Features)) similarly  for line 319

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+				if log.IsDebug() {
+					log.Debugf("Add to filter: %s", feature.Name)
+				}
+			} else {
+				var colName = feature.Name
+				var colType = "default"
```
smkniazi:
```suggestion
				var colType = api.DRT_DEFAUL
```

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+	fsResp.Features = *features
+	if fsReq.MetadataRequest != nil {
+		fsResp.Metadata = *GetFeatureMetadata(metadata, fsReq.MetadataRequest)
+	}
+	return http.StatusOK, nil
+}
+
+func checkRondbResponse(rondbResp *api.BatchResponseJSON) *feature_store.RestErrorCode {
+	for _, result := range *rondbResp.Result {
+		if *result.Code != 200 && *result.Code != 404 {
```
smkniazi:
```suggestion
   if *result.Code != http.StatusOK && *result.Code != http.StatusNotFound {                                                                      
```

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+		var fsError = TranslateRonDbError(http.StatusBadRequest, ronDbErr.Error())
+		return fsError.GetStatus(), fsError.GetError()
+	}
+	var dbResponseIntf = getPkReadResponseJSON(*metadata)
+	code, ronDbErr := h.dbBatchReader.Execute(readParams, *dbResponseIntf)
+	if ronDbErr != nil {
+		var fsError = TranslateRonDbError(code, ronDbErr.Error())
+		return fsError.GetStatus(), fsError.GetError()
+	}
+	jsonResponse := (*dbResponseIntf).String()
```
smkniazi:
```suggestion

```

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+	if log.IsDebug() {
+		log.Debugf("Rondb response: code: %d, error: %s, body: %s", code, ronDbErr, jsonResponse)
+	}
```
smkniazi:
```suggestion
	if log.IsDebug() {
        	jsonResponse := (*dbResponseIntf).String()
		log.Debugf("Rondb response: code: %d, error: %s, body: %s", code, ronDbErr, jsonResponse)
	}
```

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+	rondbResp := api.BatchResponseJSON{}
+	json.Unmarshal([]byte(jsonResponse), &rondbResp)
+	fsError := checkRondbResponse(&rondbResp)
+	if fsError != nil {
+		return fsError.GetStatus(), fsError.GetError()
+	}
```
smkniazi:
```suggestion
	   rondbResp := (*dbResponseIntf).(*api.BatchResponseJSON)                                                                                                              
```

smkniazi: marshall and Unmarsall is  VERY expensive. simply type cast the response to   api.BatchResponseJSON

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+	return nil
+}
+
+func GetFeatureMetadata(metadata *feature_store.FeatureViewMetadata, metaRequest *api.MetadataRequest) *[]*api.FeatureMeatadata {
+	featureMetadataArray := make([]*api.FeatureMeatadata, metadata.NumOfFeatures)
+	for featureKey, metadata := range metadata.PrefixFeaturesLookup {
+		featureMetadata := api.FeatureMeatadata{}
+		if metaRequest.FeatureName {
+			var fk = featureKey
```
smkniazi, line 247: why do you create new variable named fk. why not assigned the address of featureKey directly to featureMetadata.Name

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+			featureMetadata.Name = &fk
```
smkniazi, line 248:
```suggestion
     featureMetadata.Name = &featureKey                                                                                                                   
```

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+		}
+		if metaRequest.FeatureType {
+			var ft = metadata.Type
+			featureMetadata.Type = &ft
```
smkniazi, line 252:
```suggestion
	       featureMetadata.Type = &metadata.Typ
```

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+		} else {
+			fsError = feature_store.WRONG_DATA_TYPE
+		}
+
+	} else if strings.Contains(err, "Wrong number of primary-key columns.") ||
+		strings.Contains(err, "Wrong primay-key column.") ||
+		strings.Contains(err, "Column does not exist.") {
+		fsError = feature_store.INCORRECT_PRIMARY_KEY.NewMessage(err)
+	} else {
+		if code == 400 {
```
smkniazi: same as above. read the error string from the C layers

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+		}
+	}
+	return fsError
+}
+
+func GetFeatureValues(ronDbResult *[]*api.PKReadResponseWithCodeJSON, entries *map[string]*json.RawMessage, featureView *feature_store.FeatureViewMetadata) (*[]interface{}, api.FeatureStatus) {
+	featureValues := make([]interface{}, featureView.NumOfFeatures)
+	var status = api.FEATURE_STATUS_COMPLETE
+	for _, response := range *ronDbResult {
+		if *response.Code == 404 {
```
smkniazi:
```suggestion
		if *response.Code == http.StatusNotFound {
```

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+			status = api.FEATURE_STATUS_MISSING
+		} else if *response.Code != 200 {
```
smkniazi:
```suggestion
		} else if *response.Code != http.StatusOK {
```

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
smkniazi:
```suggestion
		if code == http.StatusBadRequest {
```

storage/ndb/rest-server/rest-api-server/internal/integrationtests/batchfeaturestore/handler_test.go
```diff
@@ -0,0 +1,986 @@
+/*
```
smkniazi, line 1: bit of nit-picking but it makes the code little bit more clean
https://repo.hops.works/dev/salman/batchfeaturestore.handler_test.go.patch

storage/ndb/rest-server/rest-api-server/internal/integrationtests/batchfeaturestore/performance_test.go
```diff
@@ -0,0 +1,171 @@
...
+	 This test can be run as follows:
+
+		 go test \
+			 -test.bench BenchmarkSimple \
+			 -test.run=thisexpressionwontmatchanytest \
+			 -cpu 1,2,4,8 \
+			 -benchmem \
+			 -benchtime=100x \ 		// 100 times
+			 -benchtime=10s \ 		// 10 sec
+			 ./internal/integrationtests/pkread/
```
smkniazi: please update the package name here

storage/ndb/rest-server/rest-api-server/internal/integrationtests/batchfeaturestore/performance_test.go
```diff
@@ -0,0 +1,171 @@
...
+func Benchmark_join(b *testing.B) {
+	for _, n := range []int{5, 25, 50} {
+		run(b, "fsdb001", "sample_1n2", 1, n)
+	}
+}
+
+func getSampleData(n int, fsName string, fvName string, fvVersion int) ([][]interface{}, []string, []string, error) {
+	switch fmt.Sprintf("%s|%s|%d", fsName, fvName, fvVersion) {
+	case "fsdb001|sample_1|1":
+		return fshelper.GetNSampleData("fsdb001", "sample_1_1", n)
```
smkniazi: there are constraints defined in testdb package for database names. use those constants instead of hard coding the names here. 
you can also add constants for tables names

storage/ndb/rest-server/rest-api-server/internal/integrationtests/benchmark_util.go
```diff
@@ -0,0 +1,103 @@
...
+				method,
+				url,
+				strings.NewReader((*reqs)[rand.Intn(nReq)]))
+			req.Header.Set("Content-Type", "application/json")
+			if conf.Security.APIKey.UseHopsworksAPIKeys {
+				req.Header.Set(config.API_KEY_NAME, testutils.HOPSWORKS_TEST_API_KEY)
+			}
+			requestStartTime := time.Now()
+			resp, err := httpClient.Do(req)
+			latenciesChannel <- time.Since(requestStartTime)
```
smkniazi, line 64: 200 ---> http.StatusOK

storage/ndb/rest-server/rest-api-server/internal/integrationtests/benchmark_util.go
```diff
@@ -0,0 +1,103 @@
...
+			if resp == nil || resp.StatusCode != 200 {
+				atomic.AddInt32(&numError, 1)
+				log.Infof("Error status code: %d", resp.StatusCode)
+				if err != nil {
+					log.Info(err.Error())
+				} else {
+					var respStr []byte = make([]byte, 1000)
```
smkniazi, line 71: what would happen if the response is more than 1000 bytes

storage/ndb/rest-server/rest-api-server/internal/integrationtests/feature_store/handler_test.go
```diff
@@ -0,0 +1,690 @@
+/*
```
smkniazi, line 1: same as before do use the constants defined in the testdb package for database names. you can also add more constants there for table names.

storage/ndb/rest-server/rest-api-server/internal/integrationtests/feature_store/handler_test.go
```diff
@@ -0,0 +1,690 @@
...
+	_, err := fsmetadata.GetFeatureViewMetadata("fsdb002", "NA", 1)
+	if err == nil {
+		t.Fatalf("This should fail.")
+	}
+	if !strings.Contains(err.Error(), fsmetadata.FV_NOT_EXIST.GetReason()) {
+		t.Fatalf("This should fail with error message: %s.", fsmetadata.FV_NOT_EXIST.GetReason())
+	}
+}
+
+func Test_GetFeatureVector_success(t *testing.T) {
```
smkniazi: General comment you are mixing _CammelCase_ and _snake_case_ which makes the code difficult to read.

storage/ndb/rest-server/rest-api-server/internal/integrationtests/benchmark_util.go
```diff
@@ -0,0 +1,103 @@
...
+	"time"
+
+	"hopsworks.ai/rdrs/internal/config"
+	"hopsworks.ai/rdrs/internal/log"
+	"hopsworks.ai/rdrs/internal/testutils"
+)
+
+func RunRestTemplate(b *testing.B, method, url string, reqs *[]string, numRequests int) {
+	var nReq = len(*reqs)
+	maxProcs := runtime.GOMAXPROCS(0)
```
smkniazi:
```suggestion
	maxProcs := runtime.GOMAXPROCS(0) // Calling it with "0" does not change anything, it only returns current value for GOMAXPROCS
```

storage/ndb/rest-server/rest-api-server/internal/integrationtests/feature_store/performance_test.go
```diff
@@ -0,0 +1,104 @@
...
+	 This test can be run as follows:
+
+		 go test \
+			 -test.bench BenchmarkSimple \
+			 -test.run=thisexpressionwontmatchanytest \
+			 -cpu 1,2,4,8 \
+			 -benchmem \
+			 -benchtime=100x \ 		// 100 times
+			 -benchtime=10s \ 		// 10 sec
+			 ./internal/integrationtests/pkread/
```
smkniazi: update package name in the comment

storage/ndb/rest-server/rest-api-server/internal/integrationtests/feature_store/performance_test.go
```diff
@@ -0,0 +1,104 @@
...
+*/
+
+const totalNumRequest = 100000
+
+func Benchmark(b *testing.B) {
+	run(b, "fsdb001", "sample_1", 1)
```
smkniazi: use constants for database and table names

storage/ndb/rest-server/rest-api-server/resources/testdbs/fixed/hopsworks.sql
```diff
@@ -180,6 +182,504 @@ VALUES
...
+                                 `created` timestamp NULL DEFAULT CURRENT_TIMESTAMP,
+                                 `hive_db_id` bigint(20) NOT NULL,
+                                 PRIMARY KEY (`id`),
+                                 KEY `project_id` (`project_id`),
+                                 KEY `hive_db_id` (`hive_db_id`),
+                                --  CONSTRAINT `FK_368_663` FOREIGN KEY (`hive_db_id`) REFERENCES `metastore`.`DBS` (`DB_ID`) ON DELETE CASCADE ON UPDATE NO ACTION,
+                                 CONSTRAINT `FK_883_662` FOREIGN KEY (`project_id`) REFERENCES `project` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
+) ENGINE=ndbcluster AUTO_INCREMENT=67 DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+
+-- CREATE TABLE `on_demand_feature_group` (
```
smkniazi: if you are not planning to use these tables in the future then you can remove the commented-out tables

smkniazi (changes requested):
Please fix the following warnings 

internal/feature_store/metadata.go:238:6: this value of featureStoreName is never used (SA4006)
internal/handlers/feature_store/handler.go:110:71: Printf format %s has arg #1 of wrong type *encoding/json.RawMessage (SA5009)
internal/handlers/feature_store/handler.go:314:50: should use make([]*api.PKReadParams, 0) instead (S1019)
internal/handlers/feature_store/handler.go:318:36: should use make([]api.Filter, 0) instead (S1019)
internal/handlers/feature_store/handler.go:319:40: should use make([]api.ReadColumn, 0) instead (S1019)
internal/integrationtests/batchfeaturestore/helper.go:77:13: should use String() instead of fmt.Sprintf (S1025)
internal/integrationtests/batchfeaturestore/performance_test.go:112:14: should use String() instead of fmt.Sprintf (S1025)
internal/integrationtests/benchmark_util.go:39:2: should not assign to b.N (SA3001)
internal/integrationtests/feature_store/handler_test.go:597:3: should use copy(to, from) instead of a loop (S1001)
internal/integrationtests/feature_store/helper.go:143:15: error strings should not be capitalized (ST1005)
internal/integrationtests/feature_store/helper.go:143:15: error strings should not end with punctuation or newlines (ST1005)
internal/integrationtests/feature_store/helper.go:193:6: func isRawBytesNumerical is unused (U1000)
internal/integrationtests/feature_store/helper.go:249:13: should use String() instead of fmt.Sprintf (S1025)
internal/integrationtests/feature_store/performance_test.go:98:14: should use String() instead of fmt.Sprintf (S1025)

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
kennethmhc: Seems quite hard to extract the field otherwise as their length is not fixed. And this code only runs when there is an error

storage/ndb/rest-server/rest-api-server/internal/handlers/batchfeaturestore/handler.go
kennethmhc: `Entries` is a field in `BatchFeatureStoreRequest`. But I will make it more explicit.

storage/ndb/rest-server/rest-api-server/internal/handlers/batchfeaturestore/handler.go
kennethmhc: It has been implemented but forgot to remove the comment.

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
kennethmhc: Yes they can be arrays and objects, but for simplicity, these types will not be validated but they are still supported

storage/ndb/rest-server/rest-api-server/internal/feature_store/featurestoreerrocode.go
```diff
@@ -0,0 +1,92 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+type RestErrorCode struct {
+	code    int
+	reason  string
+	status  int
+	message string
```
olapiv, line 29: Is `message` ever used?

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
kennethmhc: because address of `featureKey` will not change in each iteration of the for loop, it will end up all `featureMetadata.Name` have the same address.

storage/ndb/rest-server/rest-api-server/internal/feature_store/metadata.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+func getFeatureIndexKey(fgId int, f string) *string {
+	var featureIndexKey = GetFeatureIndexKeyByFgIndexKey(*getFeatureGroupIndexKey(fgId), f)
+	return &featureIndexKey
+}
+
+type FeatureViewMetaDataCache struct {
+	metadataCache cache.Cache
+}
```
olapiv, line 181: Could replace with:
`type FeatureViewMetaDataCache cache.Cache`

storage/ndb/rest-server/rest-api-server/internal/feature_store/featurestoreerrocode.go
```diff
@@ -0,0 +1,92 @@
...
+		"reason":  err.reason,
+		"message": err.message,
+	}
+	strBytes, jsonErr := json.MarshalIndent(m, "", "\t")
+	if jsonErr != nil {
+		return fmt.Sprintf("Failed to marshal FeatureStoreErrorResponse. Error: %s", jsonErr)
+	} else {
+		return fmt.Sprintf("%s", strBytes)
+	}
+}
```
olapiv, line 66: Instead of defining a String() function, you can do `fmt.Printf("%+v", restErr)`. The `+` in `%+v` will print both key & values of the struct. Doesn't work if the values are pointers though.

storage/ndb/rest-server/rest-api-server/internal/integrationtests/benchmark_util.go
kennethmhc: The response will be truncated. but it is just info message.

storage/ndb/rest-server/rest-api-server/internal/feature_store/featurestoreerrocode.go
olapiv: And how come we need this in addition to the `DalError`?

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+	if err2 != nil {
+		return err2.GetError()
+	}
+	return nil
+
+}
+
+func ValidatePrimaryKey(entries *map[string]*json.RawMessage, features *map[string]*feature_store.FeatureMetadata) *feature_store.RestErrorCode {
+	// Data type check of primary key will be delegated to rondb.
+	if len(*entries) == 0 {
```
olapiv, line 78: It could also be a nil map? Not a fan of pointers to maps in general though tbh.

storage/ndb/rest-server/rest-api-server/internal/integrationtests/feature_store/handler_test.go
kennethmhc: the snake case are used because it is used to identify related test case.

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+}
+
+func New(fvMetaCache *feature_store.FeatureViewMetaDataCache, apiKeyCache apikey.Cache, batchPkReadHandler batchpkread.Handler) Handler {
+	return Handler{fvMetaCache, apiKeyCache, batchPkReadHandler}
+}
+
+func (h *Handler) Validate(request interface{}) error {
+	fsReq := request.(*api.FeatureStoreRequest)
+	metadata, err := h.fvMetaCache.Get(
+		*fsReq.FeatureStoreName, *fsReq.FeatureViewName, *fsReq.FeatureViewVersion)
```
olapiv, line 57: All of these fields could be nil? If they're all simple strings, I would not recommend using pointers in the first place, unless they should be nilable. Things can crash more easily. Apart from that, you should probably add some tests where you omit some fields. (Maybe you've already done that).

https://stackoverflow.com/questions/24642311/what-is-the-point-of-passing-a-pointer-to-a-strings-in-go-golang/75958030#75958030

storage/ndb/rest-server/rest-api-server/resources/testdbs/fixed/hopsworks.sql
kennethmhc: I will need them in the upcoming PR.

storage/ndb/rest-server/rest-api-server/internal/feature_store/featurestoreerrocode.go
```diff
@@ -0,0 +1,92 @@
...
+var FETCH_METADATA_FROM_CACHE_FAIL = &RestErrorCode{9, "Fetching metadata from cache failed.", http.StatusInternalServerError, ""}
+var WRONG_DATA_TYPE = &RestErrorCode{10, "Wrong data type.", http.StatusUnsupportedMediaType, ""}
+var FEATURE_NOT_EXIST = &RestErrorCode{11, "Feature does not exist.", http.StatusNotFound, ""}
+var INCORRECT_PRIMARY_KEY = &RestErrorCode{12, "Incorrect primary key.", http.StatusBadRequest, ""}
+var INCORRECT_PASSED_FEATURE = &RestErrorCode{13, "Incorrect passed feature.", http.StatusBadRequest, ""}
+var READ_FROM_DB_FAIL = &RestErrorCode{14, "Reading from db failed.", http.StatusInternalServerError, ""}
+var NO_PRIMARY_KEY_GIVEN = &RestErrorCode{15, "No primary key is given.", http.StatusBadRequest, ""}
+var INCORRECT_FEATURE_VALUE = &RestErrorCode{16, "Incorrect feature value.", http.StatusBadRequest, ""}
+var FEATURE_STORE_NOT_SHARED = &RestErrorCode{17, "Accessing unshared feature store failed", http.StatusUnauthorized, ""}
+var READ_FROM_DB_FAIL_BAD_INPUT = &RestErrorCode{14, "Reading from db failed.", http.StatusBadRequest, ""}
```
olapiv, line 93: Generally a good idea to fix some error messages. Makes things more testable. So this is cool.

storage/ndb/rest-server/rest-api-server/internal/feature_store/featurestoreerrocode.go
```diff
@@ -0,0 +1,92 @@
...
+		return string(strBytes)
+	}
+}
+
+func (err *RestErrorCode) Error() string {
+	return err.String()
+}
+
+func (err *RestErrorCode) GetError() error {
+	return fmt.Errorf(err.Error())
```
olapiv: No need for formatting - use `errors.New()`

storage/ndb/rest-server/rest-api-server/internal/feature_store/featurestoreerrocode.go
olapiv: Afaik the pattern of using pre-defined errors could be used by the entire repo?

storage/ndb/rest-server/rest-api-server/internal/feature_store/metadata.go
```diff
@@ -0,0 +1,272 @@
...
+	FeatureGroupVersion      int
+	FeatureGroupId           int
+	Id                       int
+	Name                     string
+	Type                     string
+	Index                    int
+	Label                    bool
+	Prefix                   string
+	TransformationFunctionId int
+}
```
olapiv, line 69: If `FeatureMetadata` is only ever referenced within `FeatureViewMetadata`/`FeatureGroupFeatures`, I don't see the necessity of having all these duplicate fields. What I mean is that it could structured a bit more like a relational schema.

storage/ndb/rest-server/rest-api-server/internal/feature_store/metadata.go
```diff
@@ -0,0 +1,272 @@
...
+	"hopsworks.ai/rdrs/internal/dal"
+	"hopsworks.ai/rdrs/internal/log"
+)
+
+type FeatureViewMetadata struct {
+	FeatureStoreName     string
+	FeatureStoreId       int
+	FeatureViewName      string
+	FeatureViewId        int
+	FeatureViewVersion   int
```
olapiv, line 39: Since this is the `FeatureViewMetadata`, you can just call the fields `Name`, `Id`, `Version` and reference a FeatureStore struct.

storage/ndb/rest-server/rest-api-server/internal/handlers/batchfeaturestore/handler.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func (h *Handler) Authenticate(apiKey *string, request interface{}) error {
+	conf := config.GetAll()
+	if !conf.Security.APIKey.UseHopsworksAPIKeys {
+		return nil
+	}
+	fsReq := request.(*api.BatchFeatureStoreRequest)
+	metadata, err := h.fvMetaCache.Get(
+		*fsReq.FeatureStoreName, *fsReq.FeatureViewName, *fsReq.FeatureViewVersion)
```
olapiv, line 97: Unnecessary call to cache? We only have 1 `FeatureStoreName`? Also, nil-pointer check?

storage/ndb/rest-server/rest-api-server/internal/handlers/batchfeaturestore/handler.go
```diff
@@ -0,0 +1,207 @@
...
+	if len(*fsReq.Entries) == 0 {
+		return feature_store.NO_PRIMARY_KEY_GIVEN.GetError()
+	}
+	if len(*fsReq.PassedFeatures) != 0 && len(*fsReq.Entries) != len(*fsReq.PassedFeatures) {
+		return feature_store.INCORRECT_PASSED_FEATURE.NewMessage("Length of passed feature does not equal to that of the entries provided in the request.").GetError()
+	}
+	return nil
+}
+
+func checkStatus(fsReq *api.BatchFeatureStoreRequest, metadata *feature_store.FeatureViewMetadata, status *[]api.FeatureStatus) int {
```
olapiv: Bad function name. Or at least add a comment.

olapiv:
There are two packages called `feature_store`. Would be cool if you could change one.

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
```diff
@@ -0,0 +1,358 @@
...
+	JSON_STRING  = "STRING"
+	JSON_BOOLEAN = "BOOLEAN"
+	JSON_NIL     = "NIL"
+	JSON_OTHER   = "OTHER"
+)
+
+type Handler struct {
+	fvMetaCache   *feature_store.FeatureViewMetaDataCache
+	apiKeyCache   apikey.Cache
+	dbBatchReader batchpkread.Handler
```
olapiv, line 47: `Reader` is a bit confusing

storage/ndb/rest-server/rest-api-server/internal/handlers/batchfeaturestore/handler.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func getBatchPkReadParamsMutipleEntries(metadata *feature_store.FeatureViewMetadata, entries *[]*map[string]*json.RawMessage, status *[]api.FeatureStatus) *[]*api.PKReadParams {
+	var batchReadParams = make([]*api.PKReadParams, 0, metadata.NumOfFeatures*len(*entries))
+	for i, entry := range *entries {
+		if (*status)[i] != api.FEATURE_STATUS_ERROR {
+			for _, param := range *fshandler.GetBatchPkReadParams(metadata, entry) {
+				var oid = fmt.Sprintf("%d%s%s", i, SEQUENCE_SEPARATOR, *(*param).OperationID)
+				(*param).OperationID = &oid
+				batchReadParams = append(batchReadParams, param)
```
olapiv, line 188: Can index instead? `batchReadParams[i] = param`? Never seen `make()` with 3 parameters though.

storage/ndb/rest-server/rest-api-server/pkg/api/feature_store_data_structs.go
```diff
@@ -0,0 +1,117 @@
...
+		return string(strBytes)
+	}
+}
+
+type FeatureStoreRequest struct {
+	FeatureStoreName   *string                      `json:"featureStoreName"`
+	FeatureViewName    *string                      `json:"featureViewName"`
+	FeatureViewVersion *int                         `json:"featureViewVersion"`
+	PassedFeatures     *map[string]*json.RawMessage `json:"passedFeatures"`
+	Entries            *map[string]*json.RawMessage `json:"entries"`
```
olapiv: `Entries` is a bit generic? What is that supposed to mean?

olapiv:
@smkniazi probably already mentioned this, but the API in the README needs to updated with the Feature Store stuff.

storage/ndb/rest-server/rest-api-server/pkg/api/feature_store_data_structs.go
```diff
@@ -0,0 +1,117 @@
...
+	RequestId          *string                      `json:"requestId"`
```
olapiv: How does this relate to the OperationId?

storage/ndb/rest-server/rest-api-server/pkg/api/feature_store_data_structs.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return fmt.Sprintf("Failed to marshal FeatureStoreRequest. Error: %v", err)
+	} else {
+		return string(strBytes)
+	}
+}
+
+type BatchFeatureStoreResponse struct {
+	Features [][]interface{}     `json:"features"`
+	Metadata []*FeatureMeatadata `json:"metadata"`
```
olapiv: Spelling `FeatureMeatadata`

storage/ndb/rest-server/rest-api-server/pkg/api/feature_store_data_structs.go
```diff
@@ -0,0 +1,117 @@
...
+func (freq BatchFeatureStoreRequest) String() string {
+	strBytes, err := json.MarshalIndent(freq, "", "\t")
+	if err != nil {
+		return fmt.Sprintf("Failed to marshal FeatureStoreRequest. Error: %v", err)
+	} else {
+		return string(strBytes)
+	}
+}
+
+type FeatureStoreRequest struct {
```
olapiv, line 52: ..this is a `FeatureRequest`?

storage/ndb/rest-server/rest-api-server/pkg/api/feature_store_data_structs.go
```diff
@@ -0,0 +1,117 @@
...
+	FeatureStoreName   *string                      `json:"featureStoreName"`
+	FeatureViewName    *string                      `json:"featureViewName"`
+	FeatureViewVersion *int                         `json:"featureViewVersion"`
+	PassedFeatures     *map[string]*json.RawMessage `json:"passedFeatures"`
+	Entries            *map[string]*json.RawMessage `json:"entries"`
+	RequestId          *string                      `json:"requestId"`
+	MetadataRequest    *MetadataRequest             `json:"metadata_options"`
```
olapiv: This sounds like you're requesting metadata. Just `Metadata`?

storage/ndb/rest-server/rest-api-server/pkg/api/feature_store_data_structs.go
```diff
@@ -0,0 +1,117 @@
...
+type FeatureStatus string
+
+const (
+	FEATURE_STATUS_COMPLETE FeatureStatus = "COMPLETE"
+	FEATURE_STATUS_MISSING  FeatureStatus = "MISSING"
+	FEATURE_STATUS_ERROR    FeatureStatus = "ERROR"
+)
+
+type FeatureStoreResponse struct {
+	Features []interface{}       `json:"features"`
```
olapiv: Why is this `[]interface` and not `json.RawMessage`?

storage/ndb/rest-server/rest-api-server/pkg/api/feature_store_data_structs.go
```diff
@@ -0,0 +1,117 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package api
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+type BatchFeatureStoreRequest struct {
```
olapiv, line 25: Just add a comment what this is actually requesting. Is it requesting multiple features? Maybe name it to `BatchFeatureRequest`?

storage/ndb/rest-server/rest-api-server/pkg/api/feature_store_data_structs.go
```diff
@@ -0,0 +1,117 @@
...
+	} else {
+		return string(strBytes)
+	}
+}
+
+type FeatureStoreRequest struct {
+	FeatureStoreName   *string                      `json:"featureStoreName"`
+	FeatureViewName    *string                      `json:"featureViewName"`
+	FeatureViewVersion *int                         `json:"featureViewVersion"`
+	PassedFeatures     *map[string]*json.RawMessage `json:"passedFeatures"`
```
olapiv, line 56: Comment on what this means?

olapiv (changes requested):
I left a fair amount of comments for now, but obvs did not look at everything. Will have another look once they are addressed. No need to address absolutely everything.

Generally good though! Definitely a chunk of work 👍🏼

storage/ndb/rest-server/rest-api-server/internal/feature_store/featurestoreerrocode.go
kennethmhc: `message` is used when creating new `RestErrorCode` instance with additional error message. And `RestErrorCode` return error message in FeatureStore response format, which is different from `DalError`

storage/ndb/rest-server/rest-api-server/internal/feature_store/metadata.go
kennethmhc: FeatureViewMetaDataCache only has a `Get` method, other method from `cache.Cache` are not needed.

smkniazi (dismissed):

storage/ndb/rest-server/rest-api-server/internal/feature_store/featurestoreerrocode.go
kennethmhc: Good to know! but the purpose of `String()` here is to return response in json format.

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
kennethmhc: It is not suppose to be nil. But the pointer is coming from `Entries` in `FeatureStoreRequest`

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
kennethmhc: They are not supposed to be nil but they come from `FeatureStoreRequest`

storage/ndb/rest-server/rest-api-server/internal/feature_store/metadata.go
kennethmhc: Some metadata are used for example in `GetFeatureIndexKeyByFeature`.

storage/ndb/rest-server/rest-api-server/internal/feature_store/metadata.go
kennethmhc: I understand that it is more structure by introducing FeatureStore struct. But since there is no operation on FeatureStore struct, the current way is more simple.

storage/ndb/rest-server/rest-api-server/internal/handlers/batchfeaturestore/handler.go
kennethmhc: The metadata is used to `ValidateAPIKey` few rows after.

storage/ndb/rest-server/rest-api-server/internal/handlers/batchfeaturestore/handler.go
kennethmhc: It will introduce unnecessary counter since it is within 2 loops.

storage/ndb/rest-server/rest-api-server/pkg/api/feature_store_data_structs.go
kennethmhc: It is coming from feature store terminology basically it is primary key.

storage/ndb/rest-server/rest-api-server/pkg/api/feature_store_data_structs.go
kennethmhc: Yea, it is not necessary.

storage/ndb/rest-server/rest-api-server/pkg/api/feature_store_data_structs.go
kennethmhc: The name of the end point is `feature_store`, so `FeatureStoreRequest` is used.  And the endpoint also return metadata.

storage/ndb/rest-server/rest-api-server/pkg/api/feature_store_data_structs.go
kennethmhc: Yes, use request optional metadata.

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
olapiv: Yeah, so essentially from

```go
postParams := api.FeatureStoreRequest{}
if err := c.BindJSON(&postParams); err != nil {
	return nil, err
}
```
So if Entries does not exist in the JSON, this will be nil?

You could use GIN's `binding:"required"` required syntax, have a look at the `BatchResponseJSON` struct.

storage/ndb/rest-server/rest-api-server/internal/handlers/batchfeaturestore/handler.go
olapiv: Looks counter-intuitive because you already have `fsReq.FeatureStoreName`.

storage/ndb/rest-server/rest-api-server/pkg/api/feature_store_data_structs.go
kennethmhc: Yes it should be `json.RawMessage`. Or it should have structure similar to `BatchOpResponse`. But for now, grpc is not implement/require yet.

storage/ndb/rest-server/rest-api-server/internal/handlers/batchfeaturestore/handler.go
olapiv: Aha, now it makes sense. Cool.

storage/ndb/rest-server/rest-api-server/internal/handlers/feature_store/handler.go
olapiv: I think if you use `make()` with 3 parameters, append() won't be that critical anymore @smkniazi

storage/ndb/rest-server/rest-api-server/pkg/api/feature_store_data_structs.go
olapiv: You might have other feature_store endpoints in the future though? I.e. read?

storage/ndb/rest-server/rest-api-server/internal/feature_store/featurestoreerrocode.go
olapiv: Seems like this should be in a common.go file or so to be used by the entire GO code base. But up to you if you think otherwise.

olapiv (dismissed):

smkniazi (approved):

PR #315, opened 2023-06-07
RONDB-398: Fixes in mt.cpp for higher throughput

mronstro:
The fixes in RONDB-157 led to better responsiveness, but also led to around 50% lower throughput in cases. Thus needed to fix the m_max_send_wakeups parameters. Flush assistance was completely removed and now only send and flush assistance is performed to save CPU time.

Also fixed a problem where recv thread could go to sleep without having sent all its signals.


svenssonaxel:
Suggestions:
* `a8b751f2456 RONDB-398: Fixes in mt.cpp for higher throughput`: Clarify commit message. `Flush assistance was completely removed and now only send and flush assistance is performed to save CPU time.` could perhaps be changed to something like `Flush assistance was completely removed from execute_signals. Flushing is now performed only in mt_job_thread_main, mt_receiver_thread_main and handle_scheduling_decisions.` or similar.
* `cd7bba79788 mronstro/22.10.1 RONDB-398: Decreased delay to check receive to 1 ms`: Since this is a workaround for a possible bug still to be found, create a Jira issue for it.

svenssonaxel:
* Remove `likely` on line 7770

svenssonaxel (commented):
See suggestions above.

svenssonaxel (approved):
Caveat: I don't understand what `WITH_MEB` is or why it's removed from our build scripts. It seems to still be used in `packaging/deb-in/`.

PR #318, opened 2023-06-12
RONDB-402: API Server: Build RDRS docker image as part of build process

smkniazi:


Dockerfile.oraclelinux7
```diff
@@ -161,3 +164,38 @@ RUN --mount=type=bind,source=.,target=rondb-src \
...
 COPY --from=build-all /root/rondb-tarball .
+
+##################################
+# Intermediate Image for RDRS ####
+##################################
+
+FROM ubuntu:22.04 AS rdrs-intermediate
+
+WORKDIR /root
+COPY --from=build-all /root/rondb-tarball .
```
olapiv, line 185: You don't need to copy the tarball - you can just copy from the bin directory? Unless the bin directory is for some reason incomplete.

i.e. use `--mount=type=cache,target=rondb-bin,id=oracle7-rondb2104-bin`

Then you also don't need two separate stages.

Dockerfile.oraclelinux7
```diff
@@ -161,3 +164,38 @@ RUN --mount=type=bind,source=.,target=rondb-src \
...
+RUN mkdir -p /srv/hops/mysql/bin /srv/hops/mysql/lib /srv/hops/mysql/lib/private /srv/hops/mysql-cluster /srv/hops/mysql-cluster/log
+
+RUN --mount=type=bind,source=.,target=rondb-src \
+   mkdir /root/distro && \
+   tar xf *.tar.gz -C distro --strip-components=1 && \
+   cp distro/bin/rdrs /srv/hops/mysql/bin && \
+   cp distro/lib/libndbclient.so.6.1.0 /srv/hops/mysql/lib/ && \
+   cp distro/lib/librdrclient.so /srv/hops/mysql/lib/ && \
+   cp distro/lib/private/libcrypto.so.1.1 /srv/hops/mysql/lib/private && \
+   cp distro/lib/private/libssl.so.1.1 /srv/hops/mysql/lib/private && \
```
olapiv: Just use an env variable `MySQL_HOME_DIR` instead of retyping the PATH 8 types. Or make it the `WORKDIR`.

Dockerfile.oraclelinux7
```diff
@@ -161,3 +164,38 @@ RUN --mount=type=bind,source=.,target=rondb-src \
...
+
+##################################
+#       RDRS Image               #
+##################################
+
+FROM ubuntu:22.04 AS rdrs
+
+COPY --from=rdrs-intermediate /srv /srv
+
+ENV LD_LIBRARY_PATH=/srv/hops/mysql/lib:/srv/hops/mysql/lib/private
```
olapiv, line 213: Is this enough? You don't need e.g. 

```Dockerfile
RUN echo $LD_LIBRARY_PATH > /etc/ld.so.conf
RUN ldconfig --verbose
```
?

build_scripts/release_scripts/Jenkinsfile
```diff
@@ -80,6 +80,33 @@ pipeline {
...
+
+                cd ${WORKSPACE}/rondb
+                source ./MYSQL_VERSION
+                RONDB_VERSION="$MYSQL_VERSION_MAJOR.$MYSQL_VERSION_MINOR.$MYSQL_VERSION_PATCH"
+
+                 DOCKER_BUILDKIT=1 docker buildx build . \
+                 -f Dockerfile.oraclelinux7 \
+                 --tag rdrs:$RONDB_VERSION \
+                 --target rdrs \
+                 --build-arg BUILD_THREADS=$BUILD_CORES
```
olapiv, line 98: You want to re-use the cache? Would make sense. If not, use `--no-cache`.

olapiv (changes requested):
Generally good, I just find it irritating that you're doing this in the oraclelinux Dockerfile because the output image is ubuntu22. But it probably makes sense if you want to re-use the build cache in Jenkins. But still, I would like to see the same changes in Dockerfile.ubuntu22, so I know that they are not diverging.

Dockerfile.oraclelinux7
smkniazi: I already tried it but I ran into problems relating to versioning of lib files. Plus rdrs is not in the bin folder of `oracle7-rondb2104-bin`. It is much cleaner to rely on the tarball than on cache.

Dockerfile.oraclelinux7
smkniazi: fixed it

Dockerfile.oraclelinux7
smkniazi: It is enough.
I have also added it to /etc/ld.so.conf.d for cases where the app is run as a user other than root. 
However, there seems to be some issue with oraclelinux's ldconfig, it fails to load the /srv/hops/mysql/lib/private
It works for now as we fall back to LD_LIBRARY_PATH

build_scripts/release_scripts/Jenkinsfile
smkniazi: We should reuse the cache. Otherwise, build time will exceed 2 hrs

smkniazi (commented):
Got rid of Ubuntu base image.

olapiv:
Why is oraclelinux now an argument? Which other distributions are we expecting? `yum install` won't work on Ubuntu?

Also, can you just add the code to Dockerfile.ubuntu22 for good measure pls?

smkniazi (commented):
I had to rebase with 21.04  to deploy updated images with feature store bug fixes. I had to force push which means we have lost some comment history. I hope it is not a problem as it is a small PR

I have removed the ARG for the image name as it created some confusion. It was not the intention that users would be able to choose base images of their linking. It was more like other ARGs in the Docker file like "ARG CMAKE_VERSION=3.25.1"

Added the same changes to ubuntu file

olapiv (approved):

PR #321, opened 2023-06-15
RONDB-408: Fix bug in Dblqh when overriding ACC SCAN with TUP SCAN

svenssonaxel:


svenssonaxel:
@mronstro This is the minimal fix that I believe is appropriate for 21.04. Anything else that seems less crucial I'll similarly mark RONDB-408 but make PRs towards 22.10 only.

PR #327, opened 2023-06-21
Catch rdrs go runtime error

kennethmhc:
In general, when there are runtime errors, we should catch and log it and response with an error message.

### DIFF
--- a/storage/ndb/rest-server/rest-api-server/internal/handlers/handler.go
+++ b/storage/ndb/rest-server/rest-api-server/internal/handlers/handler.go
@@ -36,10 +36,8 @@ func Handle(h Handler, apiKey *string, request interface{}, response interface{}
 		if r := recover(); r != nil {
 			statusCode = http.StatusInternalServerError
 			err = errors.New(`{"message": "Internal server error."}`)
-			if log.IsDebug() {
-				log.Debug(r.(error).Error())
-				log.Debug(string(debug.Stack()))
-			}
+			log.Error(r.(error).Error())
+			log.Error(string(debug.Stack()))
 		}
 	}()
 	if err := h.Validate(request); err != nil {
--- a/storage/ndb/rest-server/rest-api-server/internal/handlers/handler.go
+++ b/storage/ndb/rest-server/rest-api-server/internal/handlers/handler.go
@@ -17,7 +17,13 @@
 
 package handlers
 
-import "net/http"
+import (
+	"errors"
+	"net/http"
+	"runtime/debug"
+
+	"hopsworks.ai/rdrs/internal/log"
+)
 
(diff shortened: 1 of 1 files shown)


storage/ndb/rest-server/rest-api-server/internal/handlers/handler.go
```diff
@@ -17,16 +17,29 @@
...
 }
 
-func Handle(h Handler, apiKey *string, request interface{}, response interface{}) (int, error) {
-
+func Handle(h Handler, apiKey *string, request interface{}, response interface{}) (statusCode int, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			statusCode = http.StatusInternalServerError
+			err = errors.New(`{"message": "Internal server error."}`)
+			log.Error(r.(error).Error())
```
kennethmhc: is this thread safe?

olapiv: You mean the log library? Should be, yes.

olapiv (dismissed):

olapiv:
Good idea, I think it’s important to have some kind of recovery at some stage in the code. Have you tested that the return values work if the Go code panics?

storage/ndb/rest-server/rest-api-server/internal/handlers/handler.go
```diff
@@ -17,16 +17,29 @@
...
 	Execute(request interface{}, response interface{}) (int, error)
 }
 
-func Handle(h Handler, apiKey *string, request interface{}, response interface{}) (int, error) {
-
+func Handle(h Handler, apiKey *string, request interface{}, response interface{}) (statusCode int, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			statusCode = http.StatusInternalServerError
+			err = errors.New(`{"message": "Internal server error."}`)
```
olapiv, line 38: This is our current error return format?

storage/ndb/rest-server/rest-api-server/internal/handlers/handler.go
olapiv: You can ignore the comment in the GO code regarding the thread-safe logger. What I meant there was just to use a Go logger instance/object, like zap.logger.

storage/ndb/rest-server/rest-api-server/internal/handlers/handler.go
kennethmhc: This is up for discussion.

kennethmhc:
@olapiv yes, return value works when there is a  panic in Go.

kennethmhc:
@olapiv @smkniazi please take a look

PR #333, opened 2023-06-30
Fix ndb_restor exiting with 0 despite failure

svenssonaxel:


mronstro (commented):
Given that it would be good to be able to continue the restore process even in the presence of failures (like in the App cluster now), it would be good to have a flag (not default) that could be set to not set the fatal error flag.

svenssonaxel:
Could you explain the use case? Why would we want to restore a database in an inconsistent state?

svenssonaxel:
The use case has been explained to me. I suggest a separate PR for that.

svenssonaxel:
RONDB-429 opened for `--force-restore`.

svenssonaxel:
Merged as part of https://github.com/logicalclocks/rondb/pull/352

PR #350, opened 2023-08-07
RONDB-443: Fixed compilation issues

smkniazi:


mronstro (approved):
Weird, thought I had fixed those already, but definitely ok.

PR #371, opened 2023-09-12
RONDB-475: Fix crash when combining ndb-stored users and replication

svenssonaxel:


mronstro (approved):
If it has been shown to fix the issue it looks very much correct.
So approved under the condition that it fixes the issue.
Also can be pushed as emergency fix if this is the case.
But before we can forget about the problem a test case needs to
be added in MTR for this.

PR #372, opened 2023-09-12
RONDB-473: Fix of initialisation of transaction hash in DBLQH

mronstro:
Dblqh.hpp initialised using a defined constant
DblqhInit.cpp initialised using a hardcoded value. Fixed such that DblqhInit.cpp also uses defined constant.


svenssonaxel:
There is also 
```
  m_commitAckMarkerHash.setSize(4096);
  c_firedTriggerHash.setSize(4096);
```
in DbtcInit.cpp, but I assume that is unrelated?

svenssonaxel (approved):

PR #376, opened 2023-09-15
RONDB-480: Fixed double delete issue caused by faulty merge

smkniazi:


mronstro (approved):
Isn't it a good idea to still set node_ids = nullptr and similarly for connection_string?

PR #390, opened 2023-10-13
Rondb 496: API Server. Added support for BLOB /  TEST

smkniazi:


mronstro (approved):

mronstro (approved):

storage/ndb/rest-server/rest-api-server/internal/config/init_defaults.go
```diff
@@ -47,7 +47,7 @@ func init() {
 func newWithDefaults() AllConfigs {
 	return AllConfigs{
 		Internal: Internal{
-			BufferSize:          320 * 1024,
+			BufferSize:          1024 * 1024,
```
mronstro: Why is it 1 MB here and 320 kB in previous comment?

mronstro (approved):

storage/ndb/rest-server/rest-api-server/internal/config/init_defaults.go
smkniazi: These are preallocated buffers large enough to hold any response from RonDB. With BLOBS the response buffer size is now increased to 5MB to handle the case of reading large features stored in BLOBS and TEXT.

PR #395, opened 2023-10-31
RONDB-531: New thread configuration to make better use of CPUs

mronstro:
These changesets makes it mandatory use Query Blocks in RonDB.

The Query Instances are existing in all Block threads.

The automatic thread configuration assigns half of the CPUs to
LDM threads and the rest of them to recv, TC and main threads.
The LDM, recv, TC, and main threads all contain Query Instances
and can thus handle read queries in RonDB.

The L3 cache groups can be between 4 and 16 CPUs.
The unit tests have been worked upon more and extended a bit more.
When forming the Round Robin groups from the L3 cache groups we
also ensure that Round Robin groups contain also the recv, TC and
main threads. Query instances will only assist LDM threads in the
same Round Robin group to ensure that we don't cause issues with
the CPU caches when executing those.

Introduced two new configuration variables, UseTcThreads and
UseLdmThreads that specifies whether TC threads and LDM
threads will be used by automatic thread configuration. If not
they will be executing in the recv threads.

Renamed some variables and modified some of the max number
of things related to threads, block instances and workers.

A fair amount of fixes of the portability layer for CPU configuration.


storage/ndb/src/kernel/ndbd.cpp
```diff
@@ -595,11 +595,11 @@ get_multithreaded_config(EmulatorData& ed)
...
 
   g_eventLogger->info("NDBMT: ldm_workers=%u ldm_threads=%u"
-                      " query_workers=%u query_threads=%u\n"
+                      " query_workers=%u\n"
                       " tc_threads=%u tc_workers=%u"
                       " send=%u receive=%u main_threads=%u",
                       globalData.ndbMtLqhWorkers, globalData.ndbMtLqhThreads,
-                      globalData.ndbMtQueryWorkers, globalData.ndbMtQueryThreads,
+                      globalData.ndbMtQueryWorkers,
                       globalData.ndbMtTcWorkers, globalData.ndbMtTcThreads,
```
KernelMaker: Do these two variables' positions need to be exchanged?

storage/ndb/src/common/mgmcommon/thr_config.cpp
```diff
@@ -887,32 +587,80 @@ THRConfig::do_parse(unsigned realtime,
...
+    ldm_threads +
+    tc_threads +
+    recv_threads +
+    main_threads +
+    rep_threads;
+  Uint32 calc_num_cpus = num_query_instances + send_threads;
+  require(calc_num_cpus <= num_cpus);
+
   struct ndb_hwinfo *hwinfo = Ndb_GetHWInfo(false);
   if (hwinfo->is_cpuinfo_available && num_cpus == 0)
```
KernelMaker: You reset num_cpus from zero(if it was) to a non-zero value in compute_automatic_thread_config() (Line 401), so it appears that this if condition would never be satisfied.

storage/ndb/src/kernel/vm/mt_thr_config.cpp
```diff
@@ -54,18 +54,31 @@ THRConfigApplier::find_thread(const unsigned short instancelist[], unsigned cnt)
...
   else if ((instanceNo = findBlock(DBDIH, instancelist, cnt)) >= 0)
   {
-    return &m_threads[T_MAIN][instanceNo];
+    Uint32 num_main_threads = getThreadCount(T_REP) +
+                              getThreadCount(T_MAIN);
+
+    if (num_main_threads == 2)
+      return &m_threads[T_MAIN][instanceNo];
+    else if (num_main_threads == 1)
+      return &m_threads[T_MAIN][instanceNo];
```
KernelMaker, line 77: Could it be simple like:
if (num_main_threads != 0) {
  return &m_threads[T_MAIN][instanceNo];
} else {

storage/ndb/src/kernel/vm/mt_thr_config.cpp
```diff
@@ -426,9 +426,6 @@ THRConfigApplier::do_bind(NdbThread* thread,
     case T_QUERY:
```
KernelMaker: T_QUERY is removed from THRConfigApplier::find_thread()
but it seems that T_QUERY is not removed completely:
   1     58  include/mgmcommon/thr_config.hpp <<GLOBAL>>
             T_QUERY = 9,
   2     58  src/common/mgmcommon/thr_config.cpp <<GLOBAL>>
             { "query", THRConfig::T_QUERY},
   3     83  src/common/mgmcommon/thr_config.cpp <<GLOBAL>>
             { THRConfig::T_QUERY, 0, MAX_NDBMT_QUERY_WORKERS, true, true, 0 },
   4   1356  src/common/mgmcommon/thr_config.cpp <<handle_spec>>
             type == T_QUERY ||
   5   1762  src/kernel/vm/Configuration.cpp <<setupConfiguration>>
             m_thr_config.getThreadCount(THRConfig::T_QUERY);
   6    426  src/kernel/vm/mt_thr_config.cpp <<do_bind>>
             case T_QUERY:

storage/ndb/src/kernel/vm/mt.hpp
```diff
@@ -48,10 +48,7 @@
   messages directed to other nodes and contains no blocks and
   executes thus no signals.
 */
-#define MAX_BLOCK_THREADS (MAX_MAIN_THREADS +       \
-                           MAX_NDBMT_LQH_THREADS +  \
-                           MAX_NDBMT_TC_THREADS +   \
-                           MAX_NDBMT_RECEIVE_THREADS)
+#define MAX_BLOCK_THREADS (NDBMT_MAX_BLOCK_INSTANCES)
```
KernelMaker, line 51: NDBMT_MAX_BLOCK_INSTANCES = 722;

the previous value of MAX_BLOCK_THREADS is defined as 2+332+160+80 = 574

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -1228,65 +1228,82 @@ calc_fifo_free(Uint32 ri, Uint32 wi, Uint32 sz)
...
+is_ldm_thread(unsigned thr_no)
 {
-  return is_recover_thread(thr_no);
+  if (globalData.ndbMtLqhThreads > 0)
+  {
+    return (thr_no < globalData.ndbMtLqhThreads);
+  }
+  else
+  {
+    return (thr_no < globalData.ndbMtReceiveThreads);
```
KernelMaker, line 1288: What if ndbMtTcThreads > 0?
I think this patch just makes lqh threads share recv threads, which is defined at Configuration.cpp:1831, so here need to check further to distinguish the tc threads.

**Update: no issue here. Since I found ndbMtTcThreads must be 0 if ndbMtLqhThreads is 0.**

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -2787,8 +2803,7 @@ thr_send_threads::assign_threads_to_assist_send_threads()
     thr_data *selfptr = &rep->m_thread[thr_no];
     selfptr->m_nosend = conf.do_get_nosend(selfptr->m_instance_list,
                                            selfptr->m_instance_count);
-    if (is_recover_thread(thr_no) ||
-        selfptr->m_nosend == 1)
+    if (selfptr->m_nosend == 1)
```
KernelMaker, line 2846: Why do we need two for-loops in this function? It seems that only one for-loop can work

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -4004,37 +4022,34 @@ thr_send_threads::handle_send_trp(TrpId trp_id,
...
     if (m_trp_state[trp_id].m_send_overload)        // Pause overloaded trp
     {
+      /**
+       * Go to sleep and hopefully the overload has cleared when we wake up
+       * again.
+       */
       return false;
     }
 
-    if (mt_get_send_buffer_bytes(trp_id) >= MAX_SEND_BUFFER_SIZE_TO_DELAY)
```
KernelMaker, line 4018: remove it would make the configuration MaxSendDelay ineffective

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -4004,37 +4022,34 @@ thr_send_threads::handle_send_trp(TrpId trp_id,
...
-         * will let the send thread go to sleep for as long as
-         * this trp has to wait (it is the shortest sleep we
-         * we have. For non-send threads the trp will simply
-         * be reinserted and someone will pick up later to handle
-         * things.
-         *
-         * At this point in time there are no transporters ready to
-         * send, they all are waiting for the delay to expire.
-         */
-        send_instance->m_more_trps = false;
```
KernelMaker, line 4035: remove it could make block thread continue trying to assist this delayed trp_id, which doesn't need to be assisted in this situation. it would cause unnecessary contend

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -7344,15 +7381,22 @@ static void flush_all_local_signals_and_wakeup(struct thr_data *selfptr);
...
+                            Uint32 & send_sum,
+                            Uint32 & flush_sum,
+                            bool & pending_send)
 {
+  Uint32 signal_count = 0;
   if (selfptr->m_outstanding_send_wakeups >=
       selfptr->m_max_send_wakeups)
   {
+    /* Read the prio A state often, to avoid starvation of prio A. */
+    signal_count = execute_high_prio_signals(signal,
```
KernelMaker, line 7443: **The old process of run_job_buffers is**:
if JBB is empty:
  then execute JBA and return;

Iterate JBB to get a JOB
   execute JBA again if JBA becomes non-empty;
   execute JOB;
   handle_scheduling_decisions, check if need to do_flush or do_send(no assist) to alert send thread to send at this moment;

**The new process in this patch is**:
execute JBA
if JBB is empty:
  then return;

Iterate JBB to get a JOB
  execute JOB;
  handle_scheduling_decisions, check if need to do_send(assist):
    if yes, execute JBA again and call do_send(assist);


The new process is more confusing than the old one, I guess it may want to improve the JBA's priority, but I think the old one is high enough for JBA's priority, and the new process doesn't improve further?

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -8571,7 +8560,7 @@ mt_receiver_thread_main(void *thr_arg)
     {
       watchDogCounter = 6;
       flush_all_local_signals_and_wakeup(selfptr);
-      pending_send = do_send(selfptr, true, false);
+      pending_send = do_send(selfptr, false, true);
```
KernelMaker, line 8606: The old implementation of do_send(selfptr, false, false) would still assist sending if the m_overload_status != OVERLOAD.  Why do we need to set the variable assist_send to true here? It seems to have no difference.

By the way, the first variable of assist_send_thread(), must_send is not used in the implementation. Maybe we should remove it.

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -8943,6 +8932,10 @@ handle_queue_size_stats(struct thr_data *selfptr, NDB_TICKS now)
   init_jbb_estimate(selfptr, now);
 }
 
+/**
+ * mt_job_thread_main is the function called by all block threads
+ * except the recv thread.
+ */
```
KernelMaker, line 8985: Sorry that I can only add comments on the diff part, but I have 1 question about mt_job_thread_main:
At line 9041, execute:
if (send_sum > 0) {
} else {
   pending_send = do_send(selfptr, true, true); <------------HERE
}
send_sum equals 0 here doesn't mean this block thread is not busy, because the send_sum may be reset to 0 in run_job_buffers() -> handle_scheduling_decisions(), in this situation, we should not try to assist sending.

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -9559,92 +9553,120 @@ mt_getPerformanceTimers(Uint32 self,
 const char *
 mt_getThreadDescription(Uint32 self)
```
KernelMaker: Since we already have is_XXX_thread() for each type of block thread, could we use them to simplify this function() like this:

std::string res;
if (is_recv_thread(thr_no)) {
  res += "_recv";
}
if (is_ldm_thread(thr_no)) {
  res += "_ldm";
}
if (is_tc_thread(thr_no)) {
  res += "_tc";
}
if (is_main_thread(thr_no)) {
  res += "_main";
}
if (is_rep_thread(thr_no)) {
  res += "_rep";
}
assert(!res.empty())
res.erase(res.begin());

return res;

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -11840,37 +11891,39 @@ mt_get_threads_for_blocks_no_proxy(const Uint32 blocks[],
...
-           is_ldm_thread(to) ||
-           is_recover_thread(to) ||
-           (to == from);
+    can_communicate = can_communicate ||
+                      is_tc_thread(to)   ||
+                      is_ldm_thread(to) ||
+                      (to == from);
   }
-  else if (is_recover_thread(from))
+  if (is_recv_thread(from))
```
KernelMaker, line 11981: In the previous implementation, recv thread could not communicate with other recv threads, since we combine threads together, how can we keep this limitation? Here we just loose it.

storage/ndb/src/kernel/vm/SimulatedBlock.hpp
```diff
@@ -1915,13 +1912,12 @@ class alignas(NDB_CL) SimulatedBlock :
      * is even bigger than 8 since we also want each distribute call
      * to handle a few signals before a new call is made.
      */
-    if (globalData.ndbMtQueryWorkers == 0)
-      return;
-    Uint32 num_ldm_instances = getNumLDMInstances();
+    ndbassert(globalData.ndbMtQueryWorkers > 0);
```
KernelMaker: Then, we can not support configuring query threads to 0 as usual.

storage/ndb/src/kernel/vm/SimulatedBlock.hpp
```diff
@@ -2331,16 +2298,11 @@ class alignas(NDB_CL) SimulatedBlock :
...
-    {
-      return globalData.ndbMtMainThreads +
-             globalData.ndbMtTcThreads;
-    }
-    else
-    {
-      /* Adjust for proxy instance with + 1 */
-      return globalData.ndbMtMainThreads + 1;
-    }
+    return 1;
```
KernelMaker, line 2283: Could it be 0 if lqh, tc are both 0 and recv is only 1?

storage/ndb/src/kernel/vm/Configuration.cpp
```diff
@@ -1275,7 +1273,8 @@ Uint64
 Configuration::compute_restore_memory()
 {
   Uint32 num_ldm_threads = globalData.ndbMtLqhWorkers;
-  Uint32 num_restore_threads = globalData.ndbMtRecoverThreads;
+  Uint32 num_restore_threads =
```
KernelMaker: Could we directly use globalData.ndbMtQueryWorkers to compute instead? Since recover is the same with query workers.

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -8013,13 +7989,29 @@ mt_get_instance_count(Uint32 block)
...
+ */
 void
 mt_add_thr_map(Uint32 block, Uint32 instance)
 {
   Uint32 num_lqh_threads = globalData.ndbMtLqhThreads;
   Uint32 num_tc_threads = globalData.ndbMtTcThreads;
-  Uint32 thr_no = globalData.ndbMtMainThreads;
-  Uint32 num_restore_threads = globalData.ndbMtRecoverThreads;
+  Uint32 thr_no = 0;
+  Uint32 num_lqh_workers = num_lqh_threads;
```
KernelMaker: Can we directly use globalData.ndbMtLqhWorkers instead. No need to use a local variable num_lqh_workers, check and set it again below at L8039 and L8046.

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -8067,76 +8054,46 @@ mt_add_thr_map(Uint32 block, Uint32 instance)
...
   case RESTORE:
-    if (receive_threads_only)
-    {
-      thr_no += (instance - 1);
-    }
-    else
-    {
-      thr_no += (instance - 1);
-    }
+    thr_no += (instance - 1);
```
KernelMaker, line 8100: These blocks are handled by ldm threads previously. Should we handle the situation that (ldm threads num == 0 && tc threads num > 0) here?
**Update: no issue here. Since I found ndbMtTcThreads must be 0 if ndbMtLqhThreads is 0.**

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -8067,76 +8054,46 @@ mt_add_thr_map(Uint32 block, Uint32 instance)
...
-    if (instance == num_lqh_threads + 1)
+    if (instance == num_lqh_workers + 1)
     {
       // Put extra PGMAN together with it's Proxy
       thr_no = block2ThreadId(block, 0);
     }
     else
     {
-      thr_no += (instance - 1) % num_lqh_threads;
+      thr_no += (instance - 1);
```
KernelMaker, line 8118: PGMAN block is handled by ldm threads previously. Should we handle the situation that (ldm threads num == 0 && tc threads num > 0) here?

**Update: no issue here. Since I found ndbMtTcThreads must be 0 if ndbMtLqhThreads is 0.**

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -3763,7 +3770,7 @@ check_yield(thr_data *selfptr,
         /**
          * Assist send thread once every 25 microseconds during spin.
          */
-        do_send(selfptr, true, true);
+        do_send(selfptr, false, true);
```
KernelMaker: Would assist sending here make assist too eager?

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -3930,13 +3937,24 @@ thr_send_threads::assist_send_thread(bool must_send,
...
     wakeup(&(send_instance->m_waiter_struct));
-    if (!must_send)
-    {
-      NdbMutex_Unlock(send_instance->send_thread_mutex);
-      return false;
-    }
+    NdbMutex_Unlock(send_instance->send_thread_mutex);
+    return false;
   }
+  else if (send_instance->m_num_trps_ready <= 3)
```
KernelMaker: [Just for noting]
This is an optimization to make assist sending not that eager,  but here still need to acquire send_thread_mutex which would cause lock contend. Maybe we can discuss it on workday.

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -4004,37 +4022,34 @@ thr_send_threads::handle_send_trp(TrpId trp_id,
   assert(m_trp_state[trp_id].m_thr_no_sender == NO_OWNER_THREAD);
   if (m_trp_state[trp_id].m_micros_delayed > 0)     // Trp send is delayed
   {
+    if (!is_send_thread(thr_no))
```
KernelMaker, line 4102: [Just for noting]
I think the root cause of adding this check is get_trp() could return a delayed trp, should we add a check in get_trp() instead of here?
Maybe we can discuss it on Monday.

KernelMaker:
**Note from Zhao:**

From reviewing, I think the
**Main adjustments from this patch:**

1. The number of ldm threads is configured by ThreadConfig::ldm. If configured as 0, it will reuse the recv threads, so the number of ldm threads will be the same as ThreadConfig::recv.
2. The number of tc threads is configured by ThreadConfig::tc. If configured as 0, it will reuse the recv threads, so the number of tc threads will be the same as ThreadConfig::recv.
3. The number of main threads is configured by ThreadConfig::main [0-1]. If configured as 0, it will first try to reuse the rep thread. If ThreadConfig::rep is also configured as 0, then it will reuse the first recv thread.
4. The number of rep thread is configured by ThreadConfig::rep [0-1]. If configured as 0, it will first try to reuse the main thread. If ThreadConfig::main is also configured as 0, then it will try to reuse the SECOND recv thread. If ThreadConfig::recv is just configured as 1, then it will reuse the first recv thread as the main thread does.
5. There is no need to configure ThreadConfig::query anymore; it will reuse ldm + tc + recv + main + rep threads.
6. There is no need to configure ThreadConfig::recover anymore; it will reuse query threads.

storage/ndb/src/kernel/vm/SimulatedBlock.hpp
```diff
@@ -583,8 +583,7 @@ class alignas(NDB_CL) SimulatedBlock :
   {
     return globalData.ndbMtLqhThreads +
            globalData.ndbMtTcThreads +
-           globalData.ndbMtMainThreads +
-           globalData.ndbMtRecoverThreads;
+           globalData.ndbMtMainThreads;
```
KernelMaker: main, rep threads reside behind recv threads, it seems that we shouldn't add ndbMtMainThreads here.

storage/ndb/src/kernel/vm/SimulatedBlock.hpp
```diff
@@ -1804,13 +1802,13 @@ class alignas(NDB_CL) SimulatedBlock :
...
 #define MAX_LDM_THREAD_GROUPS_PER_RR_GROUP 8
-#define MAX_RR_GROUPS ((MAX_NDBMT_QUERY_THREADS + \
+#define MAX_RR_GROUPS ((MAX_NDBMT_QUERY_WORKERS + \
                         (MIN_QUERY_INSTANCES_PER_RR_GROUP - 1)) /  \
                       MIN_QUERY_INSTANCES_PER_RR_GROUP)
 #define MAX_DISTRIBUTION_WEIGHT 16
 #define MAX_NUM_DISTR_SIGNAL \
           (MAX_DISTRIBUTION_WEIGHT * MAX_QUERY_INSTANCES_PER_RR_GROUP)
-#define MAX_DISTR_THREADS (MAX_NDBMT_LQH_THREADS)
+#define MAX_DISTR_THREADS (MAX_NDBMT_QUERY_WORKERS)
```
KernelMaker, line 1822: The number of lqh workers isn't equal with query workers all the time when ldm is not configured as 0. Here should we use MAX_NDBMT_LQH_WORKERS instead to keep the compatibility.

storage/ndb/src/kernel/ndbd.cpp
mronstro: Not sure why that should be required, can u expand on your reasoning?

storage/ndb/src/common/mgmcommon/thr_config.cpp
mronstro: Great, I noted this bug when I was benchmarking, it means I didn't get any CPU locking, great catch.

storage/ndb/src/kernel/vm/mt_thr_config.cpp
mronstro: ok, changed to that.

storage/ndb/src/kernel/vm/mt_thr_config.cpp
mronstro: ok, cleaned up and removed all references to Query threads, now only Query workers left

storage/ndb/src/kernel/ndbd.cpp
KernelMaker: the format is defined as "tc_threads = %u tc_workers = %"
so I think we should keep that order

storage/ndb/src/kernel/vm/mt.hpp
mronstro: The biggest machines at our disposal is actually 360 VCPUs, so I set the limit to more or less twice that but taking 2 main threads into account. This limit isn't really tested, so at some point in time we should probably ensure that it actually works to create such big nodes, so for the moment it is a bit of a theoretical limit.

storage/ndb/src/kernel/vm/mt.cpp
mronstro: ok, yep, if TcThreads > 0 then also LdmThreads > 0

storage/ndb/src/kernel/vm/mt.cpp
mronstro: We first round robin the sending of the LDM threads and next we round robin on all the other thread types, this is to avoid several LDM threads sharing the same send thread

storage/ndb/src/kernel/vm/mt.cpp
mronstro: Yep, MaxSendDelay have been made ineffective, it is now always on, so not configurable anymore

KernelMaker: If it is now always on, then I think we should ensure that the trp must wait for at least sendDelay before the actual sending, remove it would cause no wait anymore even if sendDelay is always on.

storage/ndb/src/kernel/vm/mt.cpp
mronstro: ok, think I tend to agree, I am working on some changes in this area, but will reintroduce that seems like a good first step.

storage/ndb/src/kernel/vm/mt.cpp
mronstro: The aim of the change is to rather decrease the prio of JBA signals, it consumes a lot of wasted CPU cycles to check for JBA signals every time one moves to a new JBB buffer, so rather check it only after executing 30-40 signals (when handle_scheduling_decisions is called).

storage/ndb/src/kernel/vm/mt.cpp
mronstro: Removed must_send from assist_send_thread. Also decreased threshold when not to assist send thread from 3 to 2. In addition I assist send thread from main and rep thread when nothing to send, we specifically wake up those threads regularly to assist send threads.

storage/ndb/src/kernel/vm/mt.cpp
mronstro: Correct finding, added if (sum == 0) to else statement

storage/ndb/src/kernel/vm/mt.cpp
mronstro: Looks like a great suggestion, implemented it.

storage/ndb/src/kernel/vm/mt.cpp
KernelMaker: I see. I think each type of thread is continuous in rep->m_thread[], so it won't be a problem to use only one for-loop.

storage/ndb/src/kernel/vm/mt.cpp
mronstro: Since the receive threads will always contain Query workers, this means that a recv thread can always send signals to other recv threads, so it is lost because it can no longer happen.

storage/ndb/src/kernel/vm/SimulatedBlock.hpp
mronstro: Correct, it is no longer possible to configure to run without query threads.

storage/ndb/src/kernel/vm/SimulatedBlock.hpp
mronstro: Nope, we always have proxy blocks, even in this case, thus it will always be 1.

storage/ndb/src/kernel/vm/Configuration.cpp
mronstro: yep, simplified the code.

storage/ndb/src/kernel/vm/mt.cpp
mronstro: Good suggestion. Implemented it.

storage/ndb/src/kernel/vm/mt.cpp
mronstro: isn't allowed to happen that num_ldm_threads == 0 AND num_tc_threads > 0

storage/ndb/src/kernel/vm/mt.cpp
mronstro: isn't allowed to happen that num_ldm_threads == 0 AND num_tc_threads > 0

storage/ndb/src/kernel/vm/mt.cpp
mronstro: I think I can reset it back to true, true, I already have safeguards inside do_send to avoid overeager behaviour.

storage/ndb/src/kernel/vm/mt.cpp
mronstro: ok

storage/ndb/src/kernel/vm/mt.cpp
mronstro: I need the distinction between having a delayed trp ready and no trp ready, so should stay as is

storage/ndb/src/kernel/vm/SimulatedBlock.hpp
mronstro: Good catch.

storage/ndb/src/kernel/vm/SimulatedBlock.hpp
mronstro: MAX_DISTR_THREADS is the number of threads that can assist in read queries (== Query workers)

KernelMaker:
I think we'd better go through the codebase to double-check all the usage of ndbMtXXXThreads for:

1. Even if we remove the independent recover threads and query threads, making them reuse other threads, can we still maintain the concept of ndbMtRecoverThreads and ndbMtQueryThreads since these two types of threads are just combined with other threads, not entirely removed? We keep these variables to maintain compatibility with MySQL upstream codes, avoiding the need to change too much code within blocks(like the modification this patch made in Dbacc, Dblqh, etc). Once we want to make them independent again or want to rebase with MySQL upstream in the future, it can be easily done. It can also help avoid unnecessary bugs due to this patch.

2. Regarding ndbMtXXXThreads and the related ndbMtXXXWorkers, I believe the introduction of ndbMtXXXWorkers may be because, once we combine threads, the count of ndbMtXXXThreads could be 0, but we still need some variables to represent the actual number of this type of threads (even if they are reusing other threads and not independent). We should carefully review all usage places of these variables in the code to ensure correctness.

For example, in this place, I'm uncertain whether it is correct or if ndbMtTcWorkers should be used instead.
image

Can we find a way to make ndbMtXXXThreads represent the actual number of threads for this type, regardless of whether they are independent or shared, to maintain compatibility with MySQL upstream code? This is crucial for avoiding unnecessary bugs and for rebasing upstream in the future.
**Anyway, this comment pertains to the patch design. But I'm not familiar with the entire rondb implementation at the moment, so it's just my suggestion for your consideration.**

storage/ndb/src/kernel/ndbd.cpp
mronstro: ok, can do that, saw the issue now :)

storage/ndb/src/kernel/vm/mt.cpp
mronstro: Possibly true, but not sure I want to rely on that, sort of hidden logic.

storage/ndb/src/kernel/vm/mt.cpp
mronstro: Yep, working on a redesign of this part. Will discuss it more in the office tomorrow.

mronstro:
It is correct that xxxThreads is the actual number of threads of a certain type and xxxWorkers is the number of threads where this type of "block thread" runs. So this means that LdmWorkers > 0, TcWorkers > 0 and recvThreads > 0 always whereas LdmThreads can be 0 and so can TcThreads. We don't have any ndbMtMainWorkers variable, this is instead calculated as being either equal to ndbMtMainThreads, if this is 0, then it is 1 if recvThreads == 1 otherwise it is 2.

Understand your concern about MySQL upstream code. However some parts of the code we will divert fairly much from the NDB design and this is one such place. Obviously any good ideas of how to make it simpler is nice. There are areas where we change very little where it makes no sense to make any major changes. mt.cpp is definitely an area where RonDB changes more than what will happen in the NDB codebase.

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -3474,8 +3539,11 @@ thr_send_threads::alert_send_thread(TrpId trp_id, NDB_TICKS now)
   struct thr_send_trps& trp_state = m_trp_state[trp_id];
 
   NdbMutex_Lock(send_instance->send_thread_mutex);
-  trp_state.m_data_available++;  // There is more to send
-  if (trp_state.m_data_available > 1)
+  Uint32 data_available = trp_state.m_data_available;  // There is more to send
+  extend_send_delay(trp_state, now);
```
KernelMaker, line 3556: It seems that everyone who calls alert_send_thread() may extend and accumulate the micros_delayed in extend_send_delay(), not only the first one who inserts the transport in the queue extends the micros_delayed.

Is it an excepted design or a bug?

storage/ndb/src/kernel/vm/mt.cpp
```diff
@@ -7627,34 +7708,13 @@ run_job_buffers(thr_data *selfptr,
   Uint32 signal_count = 0;
   Uint32 signal_count_since_last_zero_time_queue = 0;
 
+  signal_count = execute_high_prio_signals(sig,
```
KernelMaker, line 7716: In the ldm thread, send_sum and flush_sum are now always the same in this patch.
In the previous implementation, mt_job_thread_main() would perform do_flush() in every iteration and only execute do_send() once in handle_scheduling_decisions when the selfptr->m_max_signals_before_send_flush threshold was reached. 
This patch changes the frequency of do_flush(), accumulating signals in the local buffer until calling handle_scheduling_decisions() to execute do_send().

PR #421, opened 2023-12-15
RONDB-556: All block threads can become query workers

mronstro:
This patch changes the workings of query threads. To start with query threads were a physical thread, next step they were integrated with LDM threads to ensure higher throughput in writes.

In this step we extend such that all block threads, thus also recv, main, rep and TC threads can also act as query workers.

The reasoning behind this change is that e.g. a TC thread that isn't busy can assist the LDM thread by handling read queries, similarly for recv, main and rep threads.

We still want to maintain the possibility to run a thread pipeline that gives more efficient execution, but at the same time we want to use the CPUs fully since specialised threads gives some CPU usage not used since it is impossible to get a perfect balance between LDM, TC and recv thread.

Thus block threads can now both assist on sending as well as on read queries.

As part of this patch we have removed the recover threads. Since all query workers can be used for recovery the only CPUs not being used is the CPU by the send thread and this is not enough to motivate keeping recover threads.

The patch introduces two new configuration parameters: UseLdmThreads and UseTcThreads.

Setting UseLdmThreads to false means that we will run with only receive threads.

Setting UseTcThreads to false and UseLdmThreads means that we will have LDM threads and recv threads and main threads, no TC threads. This parameter is used in conjunction with AutomaticThreadConfig=1 which is the default setting.

The patch also does a major effort to improve the portability layer for RonDB. This includes handling CPUs like Intel Rapid Lake that have a mix of efficient and power CPUs. It fixes a number of bugs in the portability layer as well and extends the unit tests for this part.

Previously LDM threads was not sharing L3 cache groups with TC and recv threads. With this patch we have changed this such that also other block threads are part of L3 cache groups and thus read assistance is only provided to LDMs that are part of the same Round Robin Group (thus sharing L3 cache).

This patch removes the possibility to run RonDB without query workers. This feature has not been used in RonDB for several years. Thus it is just as well to cut away this possibility.

Some updates of the query distribution was made. A constant was introduced making it possible to schedule a bit less on TC and recv threads.

Also fixing a few bugs related to query scheduling where scans were not given enough weight and it was possible to end up in an eternal loop. Some extra protection against this possibility was added.

Also since we want to make use other than LDM threads we go directly to the round robin scheduler in some cases.

The concept of Query threads per LDM was removed. In situations with hyperthreading CPUs we still make a bit of work to schedule work on the same CPU core as the LDM thread resides on.

Updated the configuration of automatic thread configuration based on experiments performed in developing this feature.

### DIFF
--- a/storage/ndb/src/kernel/vm/SimulatedBlock.cpp
+++ b/storage/ndb/src/kernel/vm/SimulatedBlock.cpp
@@ -750,16 +750,6 @@ SimulatedBlock::setWakeupThread(Uint32 wakeup_instance)
 #endif
 }
 
-bool
-SimulatedBlock::is_recover_thread(Uint32 thr_no)
-{
-#ifdef NDBD_MULTITHREADED
-  return mt_is_recover_thread(thr_no);
-#else
-  return false;
-#endif
-}
-
 void
 SimulatedBlock::setConfMaxSendDelay(Uint32 max_send_delay)
 {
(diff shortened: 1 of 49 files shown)


storage/ndb/src/kernel/vm/SimulatedBlock.cpp
```diff
@@ -6194,9 +6205,14 @@ Uint32 SimulatedBlock::get_lqhkeyreq_ref(DistributionHandler * const handle,
       block_num = DBQLQH;
     }
   }
+  if (rr_info->m_total_weight == 0)
```
KernelMaker, line 6208: Line 6198 seems to have a problem:
```c++
query_instance_no = get_shared_ldm_instance(query_instance_no);
``` 
m_threads[][].m_shared_instance is initialized at thr_config.cpp:714, it seems like be overwritten again and again there.

PR #423, opened 2023-12-21
RONDB-562: Fix thread barrier in ndb_restore

svenssonaxel:


svenssonaxel:
This fixes an issue that `ndb_restore` doesn't use thread barrier when `--disable-indexes` is set and `--restore-meta` is unset. This means ndb_restore can change the schema in one thread while attempting to restore data in another thread. This was discovered by a test case written for RONDB-555. This PR includes no test case since it is difficult to write a test case without the code changes in RONDB-555. RONDB-555 will have a test case that covers RONDB-562 as well.

svenssonaxel:
Will be part of another PR.

PR #424, opened 2023-12-24
RONDB-555

svenssonaxel:


storage/ndb/tools/restore/restore_main.cpp
```diff
@@ -1187,6 +1185,10 @@ bool create_consumers(RestoreThreadData *data)
     // m_metadata_work_requested = 1 and m_restore_meta = 0, the thread will
     // do only the init work, and skip the ndbapi function calls to create or
     // delete the metadata objects.
+
+  if (_restore_meta)
```
mronstro, line 1189: Moving this if-statement to after the comments screw up the indentation of the comment

mronstro (approved):

mronstro (approved):

mronstro (approved):

mronstro (approved):

mronstro (approved):

mronstro (approved):

storage/ndb/tools/restore/restore_main.cpp
svenssonaxel: Yes, I thought bad indentation was better than changing all the lines, but if you prefer I'll correct the indentation.

mronstro: Actually looking at the patch the only thing changed was that you added another line of comment, so not sure why you want to change indentation

svenssonaxel: The comment block is about several command-line options, but was placed inside an if block for one of those flags. I tried to clarify by adjusting wording and moving comment out of if block.

svenssonaxel:
Most of this is in https://github.com/logicalclocks/rondb/pull/452.

PR #444, opened 2024-02-20
RONDB-616: calc_page_free_bits should return 2 page is full, not 3, a…

mronstro:
…lso important for variable sized data to use index 2


KernelMaker:
please add it back in this patch:
image

KernelMaker (approved):

PR #457, opened 2024-04-03
RONDB-592: API Key

641bill:
- Add API Key authentication between execution and validation
- Enable API Key caching
- Add GoogleTest C++ tests for API key caching
- Adjust the MTR tests accordingly

@smkniazi


smkniazi (changes requested):
[api-key-review.txt](https://github.com/user-attachments/files/15758661/api-key-review.txt)

641bill:
Regarding the pthread locks:
- It's not easier to use ReadLock class to replace the explicit pthread locks since they need to be released before the function returns.

Order of Validate and Authenticate:
- Was trying to follow Go server's order

smkniazi (dismissed):
Please create a PR for improving API cache using single or fixed number of multiple update threads

mronstro (approved):

PR #458, opened 2024-04-10
RONDB-653: Upgrade from 22.10.3 in 22.10 series should not be an issue

mronstro:


sql/dd/impl/sdi.cc
```diff
@@ -380,8 +380,13 @@ bool CheckDefaultCompatibility(const RJ_Document &doc) {
 
   const RJ_Value &mysqld_version_id = doc["mysqld_version_id"];
   assert(mysqld_version_id.IsUint64());
-  if (mysqld_version_id.GetUint64() > std::uint64_t(MYSQL_VERSION_ID)) {
+  if (mysqld_version_id.GetUint64() > std::uint64_t(MYSQL_VERSION_ID) &&
+      ((mysqld_version_id.GetUint64() < std::uint64_t(221003)) ||
+       (mysqld_version_id.GetUint64() >= std::uint64_t(221100)))) {
```
svenssonaxel: This is a little confusing. Clarify to
```
std::uint64_t running_mysql_version = std::uint64_t(MYSQL_VERSION_ID);
std::uint64_t storage_mysql_version = mysqld_version_id.GetUint64();
if (storage_mysql_version > running_mysql_version) &&
    !((std::uint64_t(221003) <= storage_mysql_version) &&
      (std::uint64_t(221003) <= running_mysql_version)) &&
      (storage_mysql_version < std::uint64_t(221100)) &&
      (running_mysql_version) < std::uint64_t(221100))))
```
or, if you'd rather have it short, shorten to
```
(mysqld_version_id.GetUint64() > std::uint64_t(MYSQL_VERSION_ID) &&
  mysqld_version_id.GetUint64() >= std::uint64_t(221100))
```

sql/dd/impl/sdi.cc
```diff
@@ -380,8 +380,13 @@ bool CheckDefaultCompatibility(const RJ_Document &doc) {
...
     // Cannot deserialize SDIs from newer versions.
+    // RonDB 22.10.3 is assumed compatible with new versions of the
```
svenssonaxel, line 386: This is a comment about *not* taking this branch, so perhaps the comment shouldn't be inside the if block. Maybe move to inside the condition?

svenssonaxel (changes requested):

sql/dd/impl/sdi.cc
mronstro: Picked the short version

sql/dd/impl/sdi.cc
mronstro: Ignored this comment

PR #472, opened 2024-06-07
RONDB-671: Add a set of new instructions to interpreter making it more complete

mronstro:


storage/ndb/include/ndbapi/NdbInterpretedCode.hpp
```diff
@@ -195,6 +275,88 @@ class NdbInterpretedCode
...
+   *   read_u16_to_reg 1 word   1 word
+   *   read_u32_to_reg 1 word   1 word
+   *   read_int64_to_reg 1 word 1 word
+   *
+   * @param RegDest Register to store operation result in
+   * @param RegSource1 Register to use as LHS of operator
+   * @param Constant to use as RHS of operator
+   * @return 0 if successful, -1 otherwise
+   */
+  int read_u8_to_reg(Uint32 RegDest, Uint16 memory_offset);
```
m-g-r: These four commands allow to copy from memory based on an immediate value or, from the perspective of the interpreted code, constant.

For me the offsets are result of a interpreted code computation, and thus they need to be register values not constants.

storage/ndb/include/kernel/Interpreter.hpp
```diff
@@ -85,18 +88,78 @@ class Interpreter {
...
+  static constexpr Uint32 READ_INT64_MEM_TO_REG = 52;
+  static constexpr Uint32 WRITE_REG_TO_MEM = 53;
+
   /**
    * Macros for creating code
    */
-  static Uint32 Read(Uint32 AttrId, Uint32 Register);
-  static Uint32 Write(Uint32 AttrId, Uint32 Register);
+  static Uint32 Read(Uint32 AttrId, Uint32 RegDest);
+  static Uint32 ReadFull(Uint32 AttrId, Uint32 RegMemOffset, Uint32 RegDest);
```
m-g-r, line 197: Here on the other hand, I guess the memory offset is almost always a constant. I just read the whole column at heap location 35000, to be able to load the whole column.

So I'd always first load-const-u32 the memory offset into a register, and then use the register in the read-full call.

Of course, it works and it gives more flexibility. That said, a variant where the memoffset is a constant is probably more often used.

m-g-r:
I have tested read-full, read-u16-to-mem and read-u32-to-mem now.
I read a varbinary column into memory at 35000 with: 
```
(load-const-u32 code 1 35000)
(read-full code vis-column 1 7)
```
afterwards I could load the length in the first 2 bytes into register 3 with:
```
(read-u16-to-reg code 3 35000)
```
So far so good.

The varbinary column contains u32 numbers then, so I tried to read the first one after the 2 bytes of the length with:
```
(read-u32-to-reg code 2 35002)
```
But that didn't work.

What strangely seems to work reliable (I tested with different columns), is to read the first number at offset 6:
```
(read-u32-to-reg code 2 35006)
```
The second at base + 2 + 2*4 = 10:
```
(read-u32-to-reg code 2 35010)
```
the third at base + 2 + 3*4 = 14:
```
(read-u32-to-reg code 2 35014)
```
Strange... all elements seem to be shifted by 4.

full test program looks like this, if that is of any help:
```
(let ((reg-memory-offset 1)
      (ordinal 2)
      (length 3)
      (value 6)
      (extra 7)
      (memory-offset 35000))
  (load-const-u32 code reg-memory-offset memory-offset)
  (read-full code vis-column reg-memory-offset extra)
  (read-u16-to-reg code length memory-offset)
  (read-u32-to-reg code ordinal 35006)
  (load-const-u32 code extra 6)
  (branch-lt code length extra 2)
  (load-const-u32 code value 2)
  (branch-ne code ordinal value 3)
  (interpret-exit-ok code)
  (def-label code 3)
  (def-label code 2)
  (interpret-exit-nok code)
  (finalise code))
```
The program selects all columns where the visibility column has a length of at least 6 (=2 bytes + 4 bytes for one u32 number), and that first number is 2. This works as shown.

storage/ndb/include/ndbapi/NdbInterpretedCode.hpp
mronstro: Ok, will ensure that these variants are also supported, haven't decided on how yet, but will fix it

storage/ndb/include/kernel/Interpreter.hpp
mronstro: Will take that into consideration,  will have to consider the space of opCodes as well

mronstro: Regarding your problem with ReadFull it is by design.
The first 4 bytes is the header, it contains the read length (lower 16 bits) and the attribute id (higher id bits)
Next comes the 2 bytes little-endian length and finally at pos 6 comes the varbinary.

storage/ndb/include/ndbapi/NdbInterpretedCode.hpp
```diff
@@ -162,9 +181,43 @@ class NdbInterpretedCode
...
+   * on e.g. VARBINARY columns. This means it isn't intended
+   * for use with columns using character sets.
+   *
+   * The memory offset in memory must be smaller than the
+   * largest column that can be read and still fit within
+   * the heap memory, thus around 35000 currently. The
+   * memory offset is in bytes and must be a multiple of
+   * 8.
+   *
+   * Important notice is that when one reads a VARBINARY(255)
```
m-g-r, line 223: Hi, thank you for your helpful replies. I suggest to add here that it is not the one or two bytes of the varbinary lenght, but also a 4 byte header as you wrote in the other comment:
```
The first 4 bytes is the header, it contains the read length (lower 16 bits) and the attribute id (higher id bits)
Next comes the 2 bytes little-endian length and finally at pos 6 comes the varbinary.
```
Thanks, Max

mronstro: Yep, planning to write a new documentation section also and invite more people to comment the design. I have added 4 new instructions of reading from memory and 1 new to write to memory using register as memory offset. I have also extended the set of opcodes to be 127 in a safe way and will add the branch on zero or non-zero that you suggested

mronstro: Also added a new move operation that uses add const

storage/ndb/include/ndbapi/NdbInterpretedCode.hpp
m-g-r: Very nice. thank you, Michael! If space is scarce you could also *replace* the existing commands with ones that use a register instead of a constant, instead of *adding* four more commands. I would expect that read_u*_to_reg with a constant offset is maybe less often? And if necessary, one could do it via a register.

m-g-r: You asked me to repeat requests in public comments, so I repeat it here. Just as per request not to annoy you. If you decide against it, that is okay, of course, and I use the replacements:

Additional branch_ne/eq_const and branch_ne/eq_zero would be helpful as well. If only one then branch_ne_const, which I need most often and could replace branch_ne_zero already.

Reason: While it is possible to load constants into registers first and then use branch_ne these are done often in our algorithms, and these are especially necessary in code paths that hit often and that can usually be decided very quickly. We have some special cases just to handle the most frequent cases very quickly without having to resort to the binary search (which handles the general case). For those test I have to compare a length value to the constants 0, 1, and 2 in succession. Here branch_ne_const would be nice, as it wouldn't need a separate command and register.

In general, I could also do the necessary code changes now and provide a commit to introduce those. Not next week, but the week after. At the same time I figure, the addition is relatively straightforward, while the decision which to add, needs more consideration.

Thanks for your time, Max

storage/ndb/include/ndbapi/NdbInterpretedCode.hpp
m-g-r: Cool, I have to work on different things next week. But will be really glad to test this out in the week after. Also I could join in and contribute some code myself if that helped. Also I am planning to write an blog article or notebook on this new feature and how to do that to push down revision determination to the data nodes using these new commands. At the very least I'll publish some examples. Cheers, Max

storage/ndb/include/ndbapi/NdbInterpretedCode.hpp
mronstro: Yep, I will add new Branch on zero/nonzero instructions as well and with the new opcodes I don't have a problem finding opcodes to use. So it won't take much time to add a few more branch instructions as well

storage/ndb/include/ndbapi/NdbInterpretedCode.hpp
mronstro: No problem if you work on other things for a week, gives me some time to develop some test cases for the
new instructions. I added a whole range of new instructions, a lot more than you asked for. Also renamed a
few of the instructions I previously added. For example I realised that just as well as reading different data
types from memory, one might want to write to memory using a different data types.

mronstro: I have found a number of new interesting use cases for the new interpreter. So I also added instructions to write to a column from interpreter using the memory and also to append to variable sized columns. Should hopefully have a finished design by early next week. The current version should be useful for you still.

m-g-r: Hi Mikael, I've had a look at the new instructions. Really great. Also writing commands will be really interesting in the future. Our import pipeline for rondb is already faster than our other back ends. But updating visibilities via interpreted code might make it even faster and simplify it at the same time (given our little lisp on top of interpreted code). We'll see. Thanks! Max

m-g-r: I just noticed that there is a new file [rondb-22.10.97-linux-glibc2.28-x86_64.tar.gz](https://repo.hops.works/master/rondb-22.10.97-linux-glibc2.28-x86_64.tar.gz) available: 17-Jun-2024 12:41.

Is that up to date with regard to the new squashed commit:
```
commit 52d090093733fd4d324ef718a34176205b457c7f
Author: Mikael Ronström <mikael.ronstrom@gmail.com>
Date:   Thu Jun 6 15:01:41 2024 +0200

    RONDB-671: Improved interpreter
```
Then, I'd just use it.

Sadly I cannot determine the version of the tar file. Also the binaries inside do not tell me more:
```
 rondb-22.10.97-linux-glibc2.28-x86_64__20240617T1241/bin# ./ndb_mgm --version
RonDB distrib RonDB-22.10.97, for Linux (x86_64)
```
Is there the git version hidden somewhere? We usually put it in our binaries somewhere... :-)

---
I also tried building rondb myself. I forgot that I actually had done this before, almost two years ago.
This now complains missing go for the new REST server:
```
CMake Error at storage/ndb/rest-server/rest-api-server/CMakeLists.txt:22 (message):
  Install Golang
```
I just installed golang-1.13 and tried again.

But apparently the also new rest-server2 also has new special needs:

```
CMake Error at storage/ndb/rest-server2/extra/drogon/CMakeLists.txt:60 (EXECUTE_PROCESS):
  EXECUTE_PROCESS given unknown argument "COMMAND_ERROR_IS_FATAL".


CMake Error at storage/ndb/rest-server2/extra/simdjson/CMakeLists.txt:43 (EXECUTE_PROCESS):
  EXECUTE_PROCESS given unknown argument "COMMAND_ERROR_IS_FATAL".


CMake Error at storage/ndb/rest-server2/server/CMakeLists.txt:22 (find_package):
  Could not find a package configuration file provided by "Drogon" with any
  of the following names:

    DrogonConfig.cmake
    drogon-config.cmake

  Add the installation prefix of "Drogon" to CMAKE_PREFIX_PATH or set
  "Drogon_DIR" to a directory containing one of the above files.  If "Drogon"
  provides a separate development package or SDK, be sure it has been
  installed.


-- Configuring incomplete, errors occurred!
```
And my ubuntu 20.04 doesn't know about any Drogon. Hm, maybe I should try to build it without the rest servers...

Thanks

mronstro: I think that binary tarballs have a reference to the Git version in the files docs/INFO_BIN and INFO_SRC.
Not an expert into those things, but I did build this based on the RONDB-671 branch I use for this development.

The rest2 is the new REST API Server that we are developing which uses C++ instead of Go.
Drogon is the C++ web server we use and we also use simdjson for parsing JSON.

Haven't seen those issues before, so not sure how to fix them, will ask someone that works on
the C++ REST API Server.

mronstro: I am developing some new instructions that I need for another project I am working on. Have finished
development mostly, but need to add the possibility to signal back to the NDB API the actions of
the interpreter for updates and writes. Also write some new test program for the last instructions I added.
Should hopefully be done early next week. But the current binary tarball should be perfectly usable for
your needs. When the final version is ready you will need to recompile your programs since I have done
some minor changes on the interpreter since the last build.

smkniazi: @m-g-r. I managed to reproduce the above build issue in Ubuntu 20.04. Updating cmake fixed the issue. We are using 3.25.1. You can also use docker to build rondb tar balls.  https://github.com/logicalclocks/rondb/blob/22.10-main/build_scripts/README.docker.build.md

storage/ndb/include/kernel/Interpreter.hpp
```diff
@@ -45,11 +47,14 @@ class Interpreter {
    * i = Instruction            -  5 Bits ( 0 - 5 ) max 63
```
svenssonaxel, line 47: Seems like an old bug in the comment. Should say 6 bits.

storage/ndb/include/kernel/Interpreter.hpp
```diff
@@ -45,11 +47,14 @@ class Interpreter {
...
    * y = Register 2             -  3 Bits ( 9 -11 ) max 7
+   * z = Register 3             -  3 Bits ( 12 -14 ) max 7
+   * w = Register 4             -  3 Bits ( 16 -18 ) max 7
    * b = Branch offset (only branches)
+   * b can also be constant, column id, ...
    *
    *           1111111111222222222233
    * 01234567890123456789012345678901
-   * iiiiiixxxyyy    bbbbbbbbbbbbbbbb
+   * iiiiiixxxyyyzzz bbbbbbbbbbbbbbbb
```
svenssonaxel, line 57: Should this be `iiiiiixxxyyyzzziwwwbbbbbbbbbbbb`?

storage/ndb/include/kernel/Interpreter.hpp
```diff
@@ -378,10 +812,11 @@ Interpreter::ExitLastOK(){
   return EXIT_OK_LAST;
 }
 
+/* Opcode is bit 0-5 and bit 15, giving a range of 0-127 */
 inline
 Uint32
 Interpreter::getOpCode(Uint32 op){
-  return op & 0x3f;
+  return (op & 0x3f) + (((op >> 15) & 1) << 6);
```
svenssonaxel, line 992: Why not `return (op & 0x3f) + ((op >> 9) & 0x40);`?

svenssonaxel: Or actually, perhaps even faster, `return op & 0x803f;` and then use constant values `0x8000` through `0x803f` for the opcodes

storage/ndb/include/kernel/Interpreter.hpp
svenssonaxel: Or actually `7 Bits ( 0 - 5, 15 )`

storage/ndb/include/ndbapi/NdbInterpretedCode.hpp
m-g-r: > I think that binary tarballs have a reference to the Git version in the files docs/INFO_BIN and INFO_SRC. Not an expert into those things, but I did build this based on the RONDB-671 branch I use for this development.
> 

Ah, thanks, yes, `docs/INFO_SRC` contains it:
```
commit: 52d090093733fd4d324ef718a34176205b457c7f
date: 2024-06-17 13:43:53 +0200
created at: 2024-06-17 12:40:20Z
branch: RONDB-671
```
good that is the new squashed commit. I base now on that. Got it work with the renamed commands but all the new commands are not available.

Makes it a bit harder that this branch BRANCH-671 seems to be a force push, because the development version cl-ndbapi now only works with a version of rondb that got replaced. Need to use hashes now in my branch names to reference to the right one, which is a bit painful.

Ok, I'll update cl-ndbapi now to that version 52d09009... oh, wait, another "forced update" of mikael/RONDB-671 and all commit squashed again. Hm, and commit: 52d090093733fd4d324ef718a34176205b457c7f disappeared from the public repositories.

Ok, I better base nothing on that version either. It is right now a bit of a painful process to recreated it. I try to use the just commited new fd30f8a8c1d2b880f2019f929cece87fc1710034. :-)

Could you please make a new build available at https://repo.hops.works/master/? I tried again to build the new rondb-21.01 but that was not successful yet.

Thanks for your work. Max

PS: Thanks to @smkniazi my server is building a new version now. I'll continue when it's done.

mronstro: Sorry about the squashes, makes it a bit easier to review the changes. However I am done with the changes now
for a while. Also the build process is running and will have a new 22.10.97 version ready in less than 2 hours.
We are going to work on the new interpreter as part of our RonSQL project. This means that the operations will
have a bit more data type handling and overflow handling to be MySQL compatible. However this development will be in another branch, so you should be safe to use this branch for now.

m-g-r: @smkniazi 
> @m-g-r. I managed to reproduce the above build issue in Ubuntu 20.04. Updating cmake fixed the issue. We are using 3.25.1. You can also use docker to build rondb tar balls. https://github.com/logicalclocks/rondb/blob/22.10-main/build_scripts/README.docker.build.md

Cool, thanks Salman Niazi! I am a bit reluctant to update cmake beyond what is part of ubuntu 20.04 as that is currently our base distribution everywhere. But trying without it and making things piece by piece wasn't successful so far.

I had to install libjsoncpp-dev, then uuid-dev, I've build drogon in `rondb/prod_build/storage/ndb/rest-server2/extra/drogon/drogon-1.8.7` by running build.sh manually, which worked. Then I had to build simdjson in `rondb/prod_build/storage/ndb/rest-server2/extra/simdjson/simdjson-3.3.0` by its builder.sh, which also worked. But in the end it still failed, so I probably really have to update cmake for building the new rondb.
I think about changing to the APT repository by Kitware, which should give me a recent cmake...

---
ok, I decided to do that for now. It got past the configure step and seem to be building rondb. Thanks again, Salman.

Max

m-g-r: > Sorry about the squashes, makes it a bit easier to review the changes. However I am done with the changes now for a while. Also the build process is running and will have a new 22.10.97 version ready in less than 2 hours. We are going to work on the new interpreter as part of our RonSQL project.

ok, thank you. I tried building with new cmake, it failed again. I went throught the errors, installed libudev-dev, libsasl2-dev, libldap2-dev, and started a new build. It got certainly harder to build it when one really wants to build it oneself.

your build will probably be done faster. :-) I'll use that.

about the squashes: I'll just publish it cl-ndbapi in the end when it works and is stable. No one will see the interim versions.

But maybe it would be nicer if the replaced branches were still available and would not vanish. If only as you seem to have done it in the past, as there is a remotes/mikael/RONDB-370 and a remotes/mikael/RONDB-370-new. :-D

> This means that the operations will have a bit more data type handling and overflow handling to be MySQL compatible. However this development will be in another branch, so you should be safe to use this branch for now.

good. thanks

Good evening for now, Max

storage/ndb/src/ndbapi/NdbInterpretedCode.cpp
```diff
@@ -281,6 +282,346 @@ NdbInterpretedCode::sub_reg(Uint32 RegDest,
                                RegSource2 % MaxReg));
 }
 
+int
+NdbInterpretedCode::lshift_reg(Uint32 RegDest,
+                               Uint32 RegSource1, Uint32 RegSource2)
+{
+  return add1(Interpreter::Lshift(RegDest % MaxReg, RegSource1 % MaxReg,
```
m-g-r: Just wanted to note that this silent modulo MaxReg (=8) is a bit dangerous.
So when one tries to use more registers, it silently is mapped to 0 to 7, with strange results.
Max

storage/ndb/include/ndbapi/NdbInterpretedCode.hpp
```diff
@@ -195,6 +313,105 @@ class NdbInterpretedCode
...
+  int write_size_mem(Uint32 RegSize, Uint32 RegOffset);
+  int read_uint8_to_reg_const(Uint32 RegDest, Uint32 memory_offset);
+  int read_uint16_to_reg_const(Uint32 RegDest, Uint32 memory_offset);
+  int read_uint32_to_reg_const(Uint32 RegDest, Uint32 memory_offset);
+  int read_int64_to_reg_const(Uint32 RegDest, Uint32 memory_offset);
+
+  int read_uint8_to_reg_reg(Uint32 RegDest, Uint32 RegOffset);
+  int read_uint16_to_reg_reg(Uint32 RegDest, Uint32 RegOffset);
+  int read_uint32_to_reg_reg(Uint32 RegDest, Uint32 RegOffset);
+  int read_int64_to_reg_reg(Uint32 RegDest, Uint32 RegOffset);
```
m-g-r, line 396: I noticed that you renamed all read and write operations from u8 to uint8 and so on.
This is a bit inconsistent as the already existant load_const_* are still called load_const_u8 and so on, and that should probably stay as it is as it is documented like that for a while and might break existing users' code.

Also the last one is load_const_u64, while this PR now adds read_int64* and write_int64* instead of unsigned integers.

It is no problem for me as I do not use it at least for now. But I wanted to comment on the inconsistencies.
Max

storage/ndb/include/ndbapi/NdbInterpretedCode.hpp
m-g-r: Hello, three updates:

1. ok, I've updated cl-ndbapi for the new branch as of fd30f8a8, added all now commands, got rid of high level commands that I do not need anymore now with move_reg, branch_*_const and so on. I wouldn't need any helping registers anymore, which is quite nice. Thanks. So right now I started with actually testing the new commands... More later.

2. To keep track of the force pushed branches, I added a fork of mronstro/rondb and logicalclocks/rondb in our company github account at https://github.com/datagraph/rondb/. It is just to keep track of the history. (For now no updates are to be expected but maybe we use it to suggest contributions in the future.)

3. I am using the rondb build that you kindly provided for fd30f8a8. At the same time I wanted to be in the position to also build releases again. My solution now is to just disable to include the rest servers by building `WITH_RDRS=0`. That is fine for me as I do not need it, and also I did not like that go would download 334 packages from various internet locations while it is building (see file `rondb/storage/ndb/rest-server/rest-api-server/go.sum`, I arrived at 334 by couting the unique URLs in the first column). I know that this is common practice of many developers nowadays. But I rather opt out here in this case and it seems you provided WITH_RDRS to allow this. :-)

(Also it still didn't build with the updated cmake. First I learned that go 1.13 was not new enough, I needed to change to golang-1.16 because of error `Cannot load io/fs: malformed module path "io/fs": missing dot in first path element` (see https://github.com/DefectDojo/godojo/issues/31 ). Luckily go 1.16 is also available in Ubuntu 20.04. But then I learned that golang 1.16 was also not enough, as I still got error `//go:build comment without // +build comment` and I probably would need at least golang 1.17 (see  https://www.reddit.com/r/golang/comments/rn25ov/encountering_gobuild_comment_without_build/?rdt=38578 ). But golang 1.17 is not available in Ubuntu 20.04. I could install go as the rondb user, but before continuing to chase this any further, I just looked if I could disable the RDRS which I then did.)

> The rest2 is the new REST API Server that we are developing which uses C++ instead of Go.
Drogon is the C++ web server we use and we also use simdjson for parsing JSON.

So will rest-server2 replace the go server in the end or will both continue to coexist?

Ok, now comes the fun and I try to use it. Hope it works. :-)
Max

mronstro: The rest-server2 will replace the Go REST API server, but first it must be fully completed and have the same functionality. A few months away still. We are planning a few new add-ons also to the C++ REST API.

mronstro: Building complex software products can always be a hassle, personally my strategy is to ensure I run on a system recent enough. I probably use WITH_RDRS=0 in some cases as well, but on Mac OS X using version 13.6 it mostly works, but definitely some effort is required to set everything up.

m-g-r: Partial success. There is still a bug in my code but it starts to work. There seem to be errors for the more complicated cases, but the simple ones work already.

There were a few pain points which took a few ours to understand. I'll write more on this tomorrow.
Bye, Max

storage/ndb/src/ndbapi/NdbInterpretedCode.cpp
m-g-r: One comment today: this turned to be a major pain point in connection with another thing:

The new ReadFull wants the memory-offset in a register. I used a constant and got all the strangest memory errors. It didn't complain, although I tried to use register 35000 as this was translated to register 0. But that register would just hold a very low number between 0 and 10.

storage/ndb/include/ndbapi/NdbInterpretedCode.hpp
mronstro: ok, sounds to good
I keep finding new things to add :) But also considering how to make interpreted programs look a bit more easy to understand.

storage/ndb/src/ndbapi/NdbInterpretedCode.cpp
mronstro: I am aware that programming the interpreter is currently a tad hard, my test program is hard to read since it is all numbers, so I am thinking about possible paths to clarify it

storage/ndb/include/ndbapi/NdbInterpretedCode.hpp
mronstro: I have added checks that registers are not out of bound, also done some internal changes to ensure
it is clear which opcodes are still free

storage/ndb/src/ndbapi/NdbInterpretedCode.cpp
m-g-r: Those BadRegister checks of your recent commit 1e03e28e are a good improvement I think! I would have noticed this problem much earlier.

well, yes, all numbers but you can store them still in c++ variables right? Makes it a bit more readable.
In Lisp I also use register names, that is variables with readable name and then the register number as their values.

Well, for me all labels and jumps are generated now, which makes it much easier to read already, then just the linear list of instructions.

storage/ndb/include/ndbapi/NdbInterpretedCode.hpp
m-g-r: Hi,

as promised some pain points when trying to use it with a bit more complex algorithms:

1. an interpreted code without (ic:interpret-exit-ok code) gives:
```
scan-operation-next-result() failed: Error with code 876: Inconsistent size of interpreter instructions
```

2. an interpreted code with a jump to the very end after the `exit-ok` such as:
```
(ndbapi.ic:branch-ne-const code length 0 0)
(ndbapi.ic:interpret-exit-ok code)
(ndbapi.ic:branch-label code 1)
(ndbapi.ic:def-label code 0)
(ndbapi.ic:interpret-exit-nok code)
(ndbapi.ic:def-label code 1)
```
fails with:
```
ndbapi.ic:finalise() failed: Error with code 4517: Bad label in branch instruction
```
even though this jump never happens. I think the real problem is that there is no instruction after the label, and so it cannot resolve that jump, right? If I just add one instruction at the very end, it works.

I can work around this by just putting a `exist-nok` at the end. These unnecessary jumps happen because the automatic translation of `IF`s...

3. that error of the last point is not unique. 4517 is use in two locations:
```
      if (label > m_number_of_labels)
      {
        m_error.code= 4517; // Bad label in branch instruction
        return -1;
      }
```
as well as
```
      if (labelOffset >= m_instructions_length)
      {
        m_error.code= 4517; // Bad label in branch instruction
        return -1;
      }
```
so it wasn't clear to me right away what the real problem was. First I made sure that all labels were defined, then I thought about how jumps are resolved. So here it is the second error.

I suggest to have two different errors. Or allow a jump to the very end which would not have this problem.

4. My decoding would still not work and I had to think a bit longer about your statement:

you wrote on Fri Jun 14 00:17:27 2024
>   Regarding your problem with ReadFull it is by design.
>   The first 4 bytes is the header, it contains the read length (lower 16 bits) and the attribute id (higher id bits)
>   Next comes the 2 bytes little-endian length and finally at pos 6 comes the varbinary.

I had to test this a bit to understand what it meant.
the read length (at position 0) = is the same as the second length (at positon 4) + 2, or put differently:

I use now two offsets to make my code simpler to read:
```
  real-memory-offset = 35000
  memory-offset = (+ real-memory-offset 4)
```
the real-memory-offset I use only for the read-full call,
for all other accesses I use memory-offset, which is just after the 4 bytes header.
It is:
```
At memory-offset, which is real-memory-offset+4, come two bytes
which give a COUNT of bytes following those two bytes.
At real-memory-offset the first two bytes are the read-length
which is COUNT + the two bytes of COUNT itself.
```

---

Having understood all this, my codes start to work. And now I have to debug it...  Oh, actually, I just fixed it. Seems to work. More on that in a separate message.

Cheers, Max

m-g-r: So, a little demo that it works. I start with this repository:
```
IMPLEMENTATION> (ndb.simple-scan::simple-scan :connection-string "localhost"
                                              :database-name "mgr°scantest"
                                              :table-name "quads"
                                              :index-name "PRIMARY" )
table: quads
columns:          GRAPH,      SUBJECT,    PREDICATE,       OBJECT
row     0:    104850320,     37518711,     37518704,     37518716,            8, #(4 0 0 0 5 0 0 0)
row     1:    104850320,     37518711,    104850321,    104850322,            4, #(5 0 0 0)
row     2:   4294967295,     37518711,     37518704,     37518712,           16, #(2 0 0 0 3 0 0 0 3 0 0 0 4 0 0 0)
row     3:   4294967295,     37518711,     37518704,    104850323,            4, #(6 0 0 0)
4
 ```
Just four statements, each with different visibility (to the right).


When I call function `ndb.interpreted-scan::revisioned-scan` with the revision ID to check as its first argument I get:
```
IMPLEMENTATION> (ndb.interpreted-scan::revisioned-scan 0 :connection-string "localhost"
                                                         :database-name "mgr°scantest"
                                                         :table-name "quads"
                                                         :index-name "PRIMARY" )
table: quads
columns:          GRAPH,      SUBJECT,    PREDICATE,       OBJECT
0
IMPLEMENTATION> (ndb.interpreted-scan::revisioned-scan 1 :connection-string "localhost"
                                                         :database-name "mgr°scantest"
                                                         :table-name "quads"
                                                         :index-name "PRIMARY" )
table: quads
columns:          GRAPH,      SUBJECT,    PREDICATE,       OBJECT
0
IMPLEMENTATION> (ndb.interpreted-scan::revisioned-scan 2 :connection-string "localhost"
                                                         :database-name "mgr°scantest"
                                                         :table-name "quads"
                                                         :index-name "PRIMARY" )
table: quads
columns:          GRAPH,      SUBJECT,    PREDICATE,       OBJECT
row     0:   4294967295,     37518711,     37518704,     37518712,           16, #(2 0 0 0 3 0 0 0 3 0 0 0 4 0 0 0)
1
IMPLEMENTATION> (ndb.interpreted-scan::revisioned-scan 3 :connection-string "localhost"
                                                         :database-name "mgr°scantest"
                                                         :table-name "quads"
                                                         :index-name "PRIMARY")
table: quads
columns:          GRAPH,      SUBJECT,    PREDICATE,       OBJECT
row     0:   4294967295,     37518711,     37518704,     37518712,           16, #(2 0 0 0 3 0 0 0 3 0 0 0 4 0 0 0)
1
IMPLEMENTATION> (ndb.interpreted-scan::revisioned-scan 4 :connection-string "localhost"
                                                         :database-name "mgr°scantest"
                                                         :table-name "quads"
                                                         :index-name "PRIMARY")
table: quads
columns:          GRAPH,      SUBJECT,    PREDICATE,       OBJECT
row     0:    104850320,     37518711,     37518704,     37518716,            8, #(4 0 0 0 5 0 0 0)
1
IMPLEMENTATION> (ndb.interpreted-scan::revisioned-scan 5 :connection-string "localhost"
                                                         :database-name "mgr°scantest"
                                                         :table-name "quads"
                                                         :index-name "PRIMARY")
table: quads
columns:          GRAPH,      SUBJECT,    PREDICATE,       OBJECT
row     0:    104850320,     37518711,    104850321,    104850322,            4, #(5 0 0 0)
1
IMPLEMENTATION> (ndb.interpreted-scan::revisioned-scan 6 :connection-string "localhost"
                                                         :database-name "mgr°scantest"
                                                         :table-name "quads"
                                                         :index-name "PRIMARY")
table: quads
columns:          GRAPH,      SUBJECT,    PREDICATE,       OBJECT
row     0:    104850320,     37518711,    104850321,    104850322,            4, #(5 0 0 0)
row     1:   4294967295,     37518711,     37518704,    104850323,            4, #(6 0 0 0)
2
IMPLEMENTATION> (ndb.interpreted-scan::revisioned-scan 7 :connection-string "localhost"
                                                         :database-name "mgr°scantest"
                                                         :table-name "quads"
                                                         :index-name "PRIMARY")
table: quads
columns:          GRAPH,      SUBJECT,    PREDICATE,       OBJECT
row     0:    104850320,     37518711,    104850321,    104850322,            4, #(5 0 0 0)
row     1:   4294967295,     37518711,     37518704,    104850323,            4, #(6 0 0 0)
2
```
and this is all correct I can assure you. :-)

The interpreted program is:
```
(ic:with-ic (:registers ((ordinal 0)
                         (length 1)
                         (test-position 2)
                         (test-ordinal 3) ;; ordinal loaded from vis-column-address at a position
                         (start 4)
                         (end 5)
                         ;; used to hold intermediate constants
                         (value 6)
                         (extra 7))
             :code code)
  (ic:load-const-u32 code ordinal ordinal/arg)
  (ic:load-const-u32 code value real-memory-offset)
  (ic:read-full code vis-column value extra)
  (ic:read-u16-to-reg/const code length memory-offset)
  (ic:rshift-const-reg code length length 2)
  (ic:load-const-null code test-position)
  (ic:load-const-null code test-ordinal) ;; ordinal loaded from vis-column-address at a position
  (ic:load-const-u32 code start 0)
  (ic:move-reg code end length)
  (ic:load-const-u32 code value 0)
  (ic:load-const-u32 code extra 0)

  (ic:cond
    ((zerop length)
     (ic:interpret-exit-ok code))
    ((= length (ic:load-const-u32 code value 1))
     (ic:read-u32-to-reg/const code test-ordinal (+ memory-offset 2))
     (ic:if (>= ordinal test-ordinal)
            (ic:interpret-exit-ok code)
            (ic:interpret-exit-nok code)))
    ((= length (ic:load-const-u32 code value 2))
     (ic:read-u32-to-reg/const code test-ordinal (+ memory-offset 2))
     (ic:when (>= ordinal test-ordinal)
       (ic:read-u32-to-reg/const code test-ordinal (+ memory-offset 6))
       (ic:when (< ordinal test-ordinal)
         (ic:interpret-exit-ok code)))
     (ic:interpret-exit-nok code))
    (t
     (ic:block binary-search
       (ic:add-reg code test-position start end)
       (ic:rshift-const-reg code test-position test-position 1)
       (ic:mul-const-reg code value test-position 4)
       (ic:add-const-reg code value value memory-offset)
       (ic:add-const-reg code value value 2) ;; add 2 to byte offset
       (ic:read-u32-to-reg/reg code test-ordinal value)
       (ic:cond
         ((< test-ordinal ordinal)
          (ic:if (= test-position start)
                 ;; position found
                 (ic:move-reg code test-position start)
                 (progn
                   (ic:move-reg code start test-position)
                   (binary-search))))
         ((> test-ordinal ordinal)
          (ic:if (= test-position start)
                 ;; position not found
                 (ic:load-const-null code test-position)
                 (progn
                   (ic:move-reg code end test-position)
                   (binary-search))))
         (t
          ;; not: (copy-reg test-position test-position :reg-help value)
          ;; just "return" test-position as is
          )))

     (ic:cond
       ((null test-position)
        (ic:interpret-exit-nok code))
       ((zerop (ic:evenp-reg code value test-position)) ;; complex argument is allowed for zerop as test
        (ic:interpret-exit-ok code))
       ((> length (ic:add-const-reg code value test-position 1)) ;; and (oddp test-position), but this is implied be the previous test already
        (ic:mul-const-reg code value value 4)
        (ic:add-const-reg code value value memory-offset)
        (ic:add-const-reg code value value 2) ;; add 2 to byte offset
        (ic:read-u32-to-reg/reg code test-ordinal value)
        (ic:if (= ordinal test-ordinal)
               ;; if the found value was a delete, but the next is an insert, then it is present
               (ic:interpret-exit-ok code)
               (ic:interpret-exit-nok code))))))
  (ic:interpret-exit-nok code))
```
that's in my little lisp on top of ndbapi's interpreted code.

I can just macro-expand it and removing the unnecessary `PROGN`s I get the bare program:
```
(let ((ordinal 0)
      (length 1)
      (test-position 2)
      (test-ordinal 3)
      (start 4)
      (end 5)
      (value 6)
      (extra 7))
  (ndbapi.ic:load-const-u32 code ordinal ordinal/arg)
  (ndbapi.ic:load-const-u32 code value real-memory-offset)
  (ndbapi.ic:read-full code vis-column value extra)
  (ndbapi.ic:read-u16-to-reg/const code length memory-offset)
  (ndbapi.ic:rshift-const-reg code length length 2)
  (ndbapi.ic:load-const-null code test-position)
  (ndbapi.ic:load-const-null code test-ordinal)
  (ndbapi.ic:load-const-u32 code start 0)
  (ndbapi.ic:move-reg code end length)
  (ndbapi.ic:load-const-u32 code value 0)
  (ndbapi.ic:load-const-u32 code extra 0)
  (ndbapi.ic:branch-ne-const code length 0 0)
  (ndbapi.ic:interpret-exit-ok code)
  (ndbapi.ic:branch-label code 1)
  (ndbapi.ic:def-label code 0)
  (ndbapi.ic:load-const-u32 code value 1)
  (ndbapi.ic:branch-ne code length value 2)
  (ndbapi.ic:read-u32-to-reg/const code test-ordinal (+ memory-offset 2))
  (ndbapi.ic:branch-lt code ordinal test-ordinal 4)
  (ndbapi.ic:interpret-exit-ok code)
  (ndbapi.ic:branch-label code 5)
  (ndbapi.ic:def-label code 4)
  (ndbapi.ic:interpret-exit-nok code)
  (ndbapi.ic:def-label code 5)
  (ndbapi.ic:branch-label code 3)
  (ndbapi.ic:def-label code 2)
  (ndbapi.ic:load-const-u32 code value 2)
  (ndbapi.ic:branch-ne code length value 6)
  (ndbapi.ic:read-u32-to-reg/const code test-ordinal (+ memory-offset 2))
  (ndbapi.ic:branch-lt code ordinal test-ordinal 8)
  (ndbapi.ic:read-u32-to-reg/const code test-ordinal (+ memory-offset 6))
  (ndbapi.ic:branch-ge code ordinal test-ordinal 9)
  (ndbapi.ic:interpret-exit-ok code)
  (ndbapi.ic:def-label code 9)
  (ndbapi.ic:def-label code 8)
  (ndbapi.ic:interpret-exit-nok code)
  (ndbapi.ic:branch-label code 7)
  (ndbapi.ic:def-label code 6)
  (ndbapi.ic:def-label code 10)
  (ndbapi.ic:add-reg code test-position start end)
  (ndbapi.ic:rshift-const-reg code test-position test-position 1)
  (ndbapi.ic:mul-const-reg code value test-position 4)
  (ndbapi.ic:add-const-reg code value value memory-offset)
  (ndbapi.ic:add-const-reg code value value 2)
  (ndbapi.ic:read-u32-to-reg/reg code test-ordinal value)
  (ndbapi.ic:branch-ge code test-ordinal ordinal 11)
  (ndbapi.ic:branch-ne code test-position start 13)
  (ndbapi.ic:move-reg code test-position start)
  (ndbapi.ic:branch-label code 14)
  (ndbapi.ic:def-label code 13)
  (ndbapi.ic:move-reg code start test-position)
  (ndbapi.ic:branch-label code 10)
  (ndbapi.ic:def-label code 14)
  (ndbapi.ic:branch-label code 12)
  (ndbapi.ic:def-label code 11)
  (ndbapi.ic:branch-le code test-ordinal ordinal 15)
  (ndbapi.ic:branch-ne code test-position start 17)
  (ndbapi.ic:load-const-null code test-position)
  (ndbapi.ic:branch-label code 18)
  (ndbapi.ic:def-label code 17)
  (ndbapi.ic:move-reg code end test-position)
  (ndbapi.ic:branch-label code 10)
  (ndbapi.ic:def-label code 18)
  (ndbapi.ic:branch-label code 16)
  (ndbapi.ic:def-label code 15)
  (ndbapi.ic:def-label code 16)
  (ndbapi.ic:def-label code 12)
  (ndbapi.ic:branch-ne-null code test-position 19)
  (ndbapi.ic:interpret-exit-nok code)
  (ndbapi.ic:branch-label code 20)
  (ndbapi.ic:def-label code 19)
  (ndbapi.ic:mod-const-reg code value test-position 2)
  (ndbapi.ic:branch-ne-const code value 0 21)
  (ndbapi.ic:interpret-exit-ok code)
  (ndbapi.ic:branch-label code 22)
  (ndbapi.ic:def-label code 21)
  (ndbapi.ic:add-const-reg code value test-position 1)
  (ndbapi.ic:branch-le code length value 23)
  (ndbapi.ic:mul-const-reg code value value 4)
  (ndbapi.ic:add-const-reg code value value memory-offset)
  (ndbapi.ic:add-const-reg code value value 2)
  (ndbapi.ic:read-u32-to-reg/reg code test-ordinal value)
  (ndbapi.ic:branch-ne code ordinal test-ordinal 25)
  (ndbapi.ic:interpret-exit-ok code)
  (ndbapi.ic:branch-label code 26)
  (ndbapi.ic:def-label code 25)
  (ndbapi.ic:interpret-exit-nok code)
  (ndbapi.ic:def-label code 26)
  (ndbapi.ic:branch-label code 24)
  (ndbapi.ic:def-label code 23)
  (ndbapi.ic:def-label code 24)
  (ndbapi.ic:def-label code 22)
  (ndbapi.ic:def-label code 20)
  (ndbapi.ic:def-label code 7)
  (ndbapi.ic:def-label code 3)
  (ndbapi.ic:def-label code 1)
  (ndbapi.ic:interpret-exit-nok code)
  (ndbapi.ic:finalise code))
```
98 instrutions. The first part is giving names to the 8 registers. I tested that this also works.

I could get rid of some initialization of registers in the beginning, but I wanted to be safe in this first version.

Cool!

From Berin, Max

Comment (m-g-r) [reply storage/ndb/include/ndbapi/NdbInterpretedCode.hpp:223]: Hm, the repeated labels that result from the compilation are not a problem right?
I'd expected the interpreter would already resolve repeated label like here:
```
  (ndbapi.ic:def-label code 23)
  (ndbapi.ic:def-label code 24)
  (ndbapi.ic:def-label code 22)
  (ndbapi.ic:def-label code 20)
  (ndbapi.ic:def-label code 7)
  (ndbapi.ic:def-label code 3)
  (ndbapi.ic:def-label code 1)
```
Or is that something that I should get rid of already?

PR #476, opened 2024-07-02
RONDB-675: Build & publish RonDB for ARM

olapiv:
https://hopsworks.atlassian.net/browse/RONDB-675

### DIFF
--- a/.github/workflows/publish_rondb.yml
+++ b/.github/workflows/publish_rondb.yml
@@ -22,9 +22,8 @@ concurrency:
 #         default: 'true'
 
 jobs:
-  publish-tarball:
-    # TODO: Run on Github runner on armchair
-    runs-on: ubuntu-latest
+  publish-arm-tarball:
+    runs-on: [self-hosted, ARM64]
     env:
       REPO_HOPSWORKS_HOSTNAME: repo.hops.works
       REPO_HOPSWORKS_USER: repo_upload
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_rondb.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Set up SSH
+        run: |
+          mkdir -p ~/.ssh
+          echo "$SSH_PRIVATE_KEY" > ~/.ssh/id_rsa
+          chmod 600 ~/.ssh/id_rsa
+          ssh-keyscan -H $REPO_HOPSWORKS_HOSTNAME >> ~/.ssh/known_hosts
```
svenssonaxel: Can we get this from any other source? E.g. hard-code it like the other vars above?

.github/workflows/publish_rondb.yml
```diff
@@ -0,0 +1,71 @@
+name: Publish RonDB Tarball
+on:
+  push:
+    branches:
```
svenssonaxel, line 4: Shouldn't this happen only on tags?

.github/workflows/publish_rondb.yml
```diff
@@ -0,0 +1,71 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Set up SSH
+        run: |
+          mkdir -p ~/.ssh
+          echo "$SSH_PRIVATE_KEY" > ~/.ssh/id_rsa
```
svenssonaxel: `id_ed25519`

.github/workflows/publish_rondb.yml
olapiv: We'll do that once we use tags

svenssonaxel (approved):

PR #526, opened 2024-09-26
RONDB-693: API Key cache rewrite

mronstro:
Major rewrite of API Key cache to have enough protection and still have a fairly high throughput.

Fixed usage of ndb_init and ndb_end, should be at beginning of main function and at end of the same. Meant removing it from init_rondb_connection and shutdown_rondb_connection.

Fixed compile issue with json_parser.cpp

Fixed API Key unit test program to properly initialise API Key cache before each test case.

Added lots of debugging for finding issues now and in the future.

Aligned build process with how RonDB programs are built.

Fixed minor issue in ronsql_dal in rdrs_dal.h.

Rewrote API Key and its test cases to use RonDB portability layer where appropriate.

### DIFF
--- a/storage/ndb/rest-server2/server/src/api_key.cpp
+++ b/storage/ndb/rest-server2/server/src/api_key.cpp
@@ -26,19 +26,101 @@
 #include <iostream>
 #include <memory>
 #include <openssl/evp.h>
-#include <pthread.h>
 #include <thread>
 #include <unordered_map>
 #include <functional>
 #include <algorithm>
+#include <NdbThread.h>
+#include <util/require.h>
+#include <EventLogger.hpp>
 
-std::shared_ptr<APIKeyCache> apiKeyCache;
+//#define DEBUG_AUTH 1
+//#define DEBUG_AUTH_THREAD 1
+//#define DEBUG_AUTH_TIME 1
+//#define DEBUG_AUTH_DBS 1
+
+#ifdef DEBUG_AUTH
+#define DEB_AUTH(arglist) do { g_eventLogger->info arglist ; } while (0)
+#else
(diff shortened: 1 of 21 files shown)


storage/ndb/rest-server2/server/src/api_key.cpp
```diff
@@ -26,371 +26,497 @@
...
+
+  /* Wait for all objects to complete cleanup */
+  for (int i = 0; i < NUM_API_KEY_CACHES; i++) {
+    NdbMutex_Lock(m_rwLock[i]);
+    while (m_key_cache[i].size() > 0) {
+#ifdef DEBUG_AUTH
+      Uint32 key_cache_size = (Uint32)m_key_cache[i].size();
+#endif
+      NdbMutex_Unlock(m_rwLock[i]);
+      DEB_AUTH(("m_key_cache[%d].size() = %u", i, key_cache_size));
```
smkniazi, line 113: why not 
```
#ifdef DEBUG_AUTH
      Uint32 key_cache_size = (Uint32)m_key_cache[i].size();
      DEB_AUTH(("m_key_cache[%d].size() = %u", i, key_cache_size));
#endif
      NdbMutex_Unlock(m_rwLock[i]);

```

smkniazi (approved):
Cpp unit test binary is in bin folder. Does it become part of install tarball?

storage/ndb/rest-server2/server/src/api_key.cpp
mronstro: Not sure, but pretty sure it will be part of install tarball, above is to avoid holding the mutex while printing, but obviously not important when running in debug, so can change if need be

PR #536, opened 2024-10-10
RONDB-746: Add a thread context to ensure that each thread uses its own Ndb object

mronstro:


storage/ndb/rest-server2/server/src/rdrs_rondb_connection_pool.cpp
```diff
@@ -24,16 +24,48 @@
 
 #include <util/require.h>
 
+ThreadContext::ThreadContext() {
+  m_thread_context_mutex = NdbMutex_Create();
+  require(m_thread_context_mutex);
+  m_is_shutdown = false;
```
svenssonaxel, line 49: These three assignments could be done before the body e.g. `() : m_is_shutdown(false) ... {`

storage/ndb/rest-server2/server/src/rdrs_rondb_connection_pool.cpp
```diff
@@ -91,16 +129,61 @@ RS_Status RDRSRonDBConnectionPool::AddMetaConnections(
...
+     * This is by far the most common path through the code. We are not
+     * shutting down, the Ndb object isn't in use and it exists. We only
+     * allocate the Ndb object normally on the first interaction.
+     */
+    *ndb_object = thread_context->m_ndb_object;
+    thread_context->m_is_ndb_object_in_use = true;
+    NdbMutex_Unlock(m_thread_context[threadIndex]->m_thread_context_mutex);
+    return RS_OK;
+  }
+  require(thread_context->m_is_ndb_object_in_use == false);
```
svenssonaxel, line 221: If this is correct then why not move it to the top of the function?

svenssonaxel:
Do I understand correctly that an NDB object represents a connection, and each NDB object would allocate one node id? If so, requiring one NDB object per thread would mean requiring one node id per thread. Is that necessary? Wouldn't it risk exhausting the node ids?

storage/ndb/rest-server2/server/src/rdrs_rondb_connection_pool.cpp
mronstro: One could do so, yep

storage/ndb/rest-server2/server/src/rdrs_rondb_connection_pool.cpp
mronstro: Yep, but in my view this variant is easier to read, the assignment before the function simply introduces syntactical sugar that in my opinion has no real purpose, I am sure someone disagrees :)

PR #538, opened 2024-10-15
RONDB: 717, RONDB-718: Ensure API Key for ping and health only requir…

mronstro:
…ed when API keys in general are required plus a config variable enabling ping and stat to use a configuration variable


svenssonaxel (commented):
.

PR #540, opened 2024-10-17
[RONDB-749] rdrs: ArenaMalloc

svenssonaxel:


mronstro (approved):
Not partticularly happy with the #     ifdef XXXX, all the other code uses #ifdef XXXX
Will let it go for now.

PR #554, opened 2024-11-08
RONDB-753 TTL feature [Patch 4 - feature]

KernelMaker:
Support changing TTL by 'ALTER TABLE' statement


mronstro (changes requested):
You need to use
m_ldm_instance_used->c_lqh in is_ttl_table

c_lqh refers to DBQLQH when running in query thread and this has no table info

mronstro (approved):
I noticed that the only objects you use in Dblqh are table array size, table array pointer and these are setup on every use of a query thread, but if the function ever is changed to touch any block data in DBLQH it will require the m_ldm_instance_used to call DBLQH rather than DBQLQH.

PR #556, opened 2024-11-08
RONDB-767: Add ARM-build dispatch in CI

olapiv:
https://hopsworks.atlassian.net/browse/RONDB-767


mronstro (approved):
Didn't you need to change the branches? or does it work simply because it is started manually?

olapiv:
>  or does it work simply because it is started manually?

Yes, it is unrelated to the branch. The branch just has to contain the file.

PR #585, opened 2024-12-02
RONDB-789: Automatic memory management when running in a container

mronstro:


storage/ndb/src/common/portlib/NdbHW.cpp
```diff
@@ -2147,6 +2149,20 @@ static int
 get_meminfo(struct ndb_hwinfo *hwinfo)
 {
   char buf[1024];
+  FILE * cgroup_meminfo = fopen("/sys/fs/cgroup/memory.max", "r");
+  if (cgroup_meminfo != nullptr) {
+    hwinfo->is_running_in_container = 1;
+    FileGuard g(cgroup_meminfo); // close at end...
+    int ret_code = fgets(buf, sizeof(buf), cgroup_meminfo))
+    Uint64 memory_size = 0;
+    if (sscanf(buf, "%llu", &memory_size) == 1) {
```
svenssonaxel, line 2159: This will/should fail when encountering the literal word `max`.

storage/ndb/src/common/portlib/NdbHW.cpp
```diff
@@ -2719,6 +2735,7 @@ static int Ndb_ReloadHWInfo(struct ndb_hwinfo * hwinfo)
   hwinfo->cpu_cnt_max = ncpu;
   hwinfo->cpu_cnt = ncpu;
   check_cpu_online(hwinfo);
+  hwinfo->is_running_in_container = 0;
```
svenssonaxel, line 2739: Is `Ndb_ReloadHWInfo` supposed to reset but not "get" the data? If so, then why do we call `check_cpu_online` that alters cpu info? If not, then we don't we call `get_meminfo`?

Also, why is this added to only one of the 6 implementations of `Ndb_ReloadHWInfo`?

storage/ndb/src/kernel/vm/Configuration.cpp
```diff
@@ -358,9 +358,11 @@ Configuration::get_total_memory(const ndb_mgm_configuration_iterator *p,
     struct ndb_hwinfo *hwinfo = Ndb_GetHWInfo(false);
     return hwinfo->hw_memory_size;
     total_memory_set = false;
+    globalData.theIsRunningInContainerFlag = hwinfo->is_running_in_container;
   }
   else
   {
+    globalData.theIsRunningInContainerFlag = false;
```
svenssonaxel: It seems you want `globalData.theIsRunningInContainerFlag` to be true if memory management should use container limit, and this can be false even if we run in a container. If correct, then this is a deceptive variable name and should be changed.

svenssonaxel (changes requested):

svenssonaxel (approved):

PR #589, opened 2024-12-04
Bump version 22.10.7

smkniazi:


mronstro (approved):
Don't forget to tag 22.10.7 and updating the patch number to 22.10.8 after building 22.10.7.

PR #612, opened 2024-12-26
RONDB-826: Added enable-md2 for building OpenSSL from source

mronstro:


Dockerfile.oraclelinux9
```diff
@@ -76,6 +76,8 @@ RUN mkdir $DOWNLOADS_CACHE_DIR
 # it in $OPENSSL_ROOT. We will not overwrite the system's OpenSSL library, 
 # but just install the specified version.
 # Commands are from https://linuxpip.org/install-openssl-linux/
+# enable-md2 is required in Rocky Linux 9, so should be used here,
+# some parts of the code in new REST server uses deprecated SSL features.
```
svenssonaxel, line 80: Perhaps then we should remove that usage rather than require a special and insecure build of OpenSSL?

svenssonaxel (commented):
.

PR #623, opened 2025-01-20
[WIP] complex_features_optimized

smkniazi:


mronstro (commented):
Looks good, would it be possible to avoid multi-threading it when number of complex features is 1 to avoid adding latency to those cases where only one complex feature is desired?

PR #631, opened 2025-01-29
RONDB-852 Crash in executeTrigger when an update causes a disk page e…

KernelMaker:
…xpansion.

Description:
  When using a table with variable-size disk columns, if an update to a
  disk column generates a new page and the binlog is enabled,
  executeTrigger incorrectly accesses the original page, leading to a
  crash.


mysql-test/suite/ndb/t/bug_rondb-852.test
```diff
@@ -0,0 +1,33 @@
...
+CREATE LOGFILE GROUP lg_1
+    ADD UNDOFILE 'undo_1.log'
+    INITIAL_SIZE 16M
+    UNDO_BUFFER_SIZE 2M
+    ENGINE NDBCLUSTER;
+
+CREATE TABLESPACE ts_1
+    ADD DATAFILE 'data_1.dat'
+    USE LOGFILE GROUP lg_1
+    INITIAL_SIZE 32M
```
mronstro, line 15: To ensure that test case works I think you need to add one more line:
EXTENT_SIZE 32k
to the CREATE TABLESPACE

storage/ndb/src/kernel/blocks/dbtup/DbtupTrigger.cpp
```diff
@@ -1805,7 +1812,27 @@ bool Dbtup::readTriggerInfo(TupTriggerData *const trigPtr,
         get_copy_tuple(&req_struct->prevOpPtr.p->m_copy_tuple_location);
   }
 
-  if (regTabPtr->need_expand(disk)) prepare_read(req_struct, regTabPtr, disk);
```
mronstro, line 1808: After considering it in a deep way for a few hours I think I can conclude that this is never needed.
The code here is about reading the AFTER value, for DELETEs it only reads the PK, so no need
to worry about disk pages in this case. If an extra disk page exists, then it means that the transaction
has updated the disk part. This means that the AFTER value for the disk part is in the copy row.
Thus we will not make use m_disk_page_ptr in this case.

storage/ndb/src/kernel/blocks/dbtup/DbtupTrigger.cpp
```diff
@@ -1909,7 +1936,27 @@ bool Dbtup::readTriggerInfo(TupTriggerData *const trigPtr,
           get_copy_tuple(&req_struct->prevOpPtr.p->m_copy_tuple_location);
     }
 
-    if (regTabPtr->need_expand(disk)) prepare_read(req_struct, regTabPtr, disk);
+    if (regTabPtr->need_expand(disk)) {
```
mronstro, line 1972: This code is required, but given that the above code is removed, this means that there is no TODO left.
The 2 prepare_read calls are one for after value and one for before value.

The only case where we need worry about extra disk page is with SUBSCRIPTION_BEFORE, thus with
the binlog case. And here the below code is required.

mysql-test/suite/ndb/t/bug_rondb-852.test
```diff
@@ -0,0 +1,33 @@
...
+  col_a int NOT NULL,
+  col_b varchar(16383) NOT NULL STORAGE DISK STORAGE DISK,
+  PRIMARY KEY(col_a)
+) ENGINE=ndbcluster CHARSET=latin1 TABLESPACE ts_1 STORAGE DISK PARTITION BY KEY(col_a) PARTITIONS 1;
+
+INSERT INTO t1 VALUES (1, REPEAT('a', 8192));
+INSERT INTO t1 VALUES (2, REPEAT('a', 8192));
+INSERT INTO t1 VALUES (3, REPEAT('a', 6));
+
+UPDATE t1 SET col_b = REPEAT('a', 16380) WHERE col_a = 3;
```
mronstro, line 44: This test case is good, but what happens if the bug exists, that is not clear to me.
Another test case that would be good to have is similar but the UPDATE is replaced by
BEGIN
UPDATE same way
DELETE row
COMMIT

storage/ndb/src/kernel/blocks/dbtup/DbtupTrigger.cpp
mronstro: Also don't forget to update Copyright headers of files changed

mronstro (approved):

PR #639, opened 2025-02-11
Improvements to ACTIVATE and DEACTIVATE commands

mronstro:


storage/ndb/src/mgmclient/CommandInterpreter.cpp
```diff
@@ -3074,32 +3073,34 @@ CommandInterpreter::executeDeactivate(int processId,
   iter.get(CFG_NODE_ACTIVE, &is_active);
   if (!is_active)
   {
-    ndbout_c("Node %d is already deactivated, need not be deactivated",
+    ndbout_c("Node %d is already deactivated, will deactivate still",
```
KernelMaker, line 3076: Here, if you allow the deactivate command to be executed repeatedly, regardless of the current state, it seems that the same modification is needed on the data node as well—specifically in Qmgr::execDEACTIVATE_REQ().

mysql-test/suite/ndb/t/ndb_activate.test
```diff
@@ -0,0 +1,19 @@
...
+--exec $NDB_MGM -e "16 DEACTIVATE"
+--exec $NDB_MGM -e "16 ACTIVATE"
+
+--exec $NDB_MGM -e "1 RESTART -n"
+--exec $NDB_WAITER --nowait-nodes=2 --not-started
+--exec $NDB_MGM -e "1 ACTIVATE"
+--exec $NDB_MGM -e "1 START"
+--exec $NDB_WAITER
+--echo Wait for mysqld to connect to the cluster
+--source include/ndb_not_readonly.inc
```
KernelMaker, line 19: Please also add a case on executing DEACTIVATE repeatedly

mronstro: It wasn't possible to use DEACTIVATE on data nodes in MTR, had to do that test manually, only API nodes can be executed in MTR repeatedly

storage/ndb/src/mgmclient/CommandInterpreter.cpp
mronstro: Nope, QMGR already supports repeated calls to DEACTIVATE

KernelMaker (approved):

PR #647, opened 2025-03-04
RONDB-860: Ensure that pk-read is part of relativeUrl

mronstro:


storage/ndb/rest-server2/server/src/json_parser.cpp
```diff
@@ -1457,17 +1458,26 @@ RS_Status extract_db_and_table(const std::string_view &relativeUrl,
...
+    request_type = relativeUrl.substr(lastSlashPos);
   } else {
     return CRS_Status(static_cast<HTTP_CODE>(
       drogon::HttpStatusCode::k400BadRequest),
-      ERROR_INVALID_RELATIVE_URL, std::string(rdrsErrorMessage(ERROR_INVALID_RELATIVE_URL))).status;
+      ERROR_INVALID_RELATIVE_URL,
+      std::string(rdrsErrorMessage(ERROR_INVALID_RELATIVE_URL))).status;
   }
-  return CRS_Status::SUCCESS.status;
+  if (request_type.length() == strlen(PKREAD) &&
```
KernelMaker, line 1495: This if statement doesn't have an effect

KernelMaker:
**Fixing relative-url Issues:**
Two issues still remain:

1. The relative-url parser seems to only remove one extra / at the end. If the URL has more than one trailing /, it treats it as a failure.
2. In batch pk-read, each operation's result should be processed separately. However, with this patch, if even one operation contains an invalid relative-url, the entire batch pk-read request fails.

**Fixing errno Handling:**
I'm not sure if introducing such a large modification (60+ files) is necessary for handling errno. The patch explicitly resets errno = 0 at the beginning of each thread, even for threads that don’t interact with errno at all. This seems excessive, and I’m unsure if it could cause unintended issues in the future.

Suggested Approach:
We should keep the fix as simple as possible by modifying only the places where strtol is used. A more reasonable solution would be to check for errno based on the second parameter of strtol:

```
errno = 0;
char *end;
long value = strtol(str, &end, 10);
if (end > str && *end == '\0') {
    // Success
} else {
    // Handle error
    errno;
}
```
BTW, I also double-checked how MySQL and Redis handle strtol(). In most cases, they simply use it without checking, and only in a few places do they implement checks similar to the approach above.

PR #654, opened 2025-03-14
Merge 22.10-dev into 24.10-main

smkniazi:


mronstro (changes requested):
I gather that changes to test programs in rest-server need to be applied to rest-server2 as well

smkniazi:
I do not think the changes in unit tests need to be migrated to rdr2. We both fixed the same thing differently

mronstro (approved):

PR #677, opened 2025-04-11
[RONDB-706] RDRS and ndbapi fixes

svenssonaxel:


mronstro (dismissed):
Still using system::chrono and similar stuff, use RonDB variants instead

svenssonaxel:
@mronstro now using NdbTick and NdbSleep.

smkniazi (approved):
Look good. I did not fully understand the purpose of check magic fn.

svenssonaxel:
> Look good. I did not fully understand the purpose of check magic fn.

The `magic` member variable is set to a certain random-looking constant at object construction and to 0 at destruction. It is used to guard against memory bugs that e.g. mess up object life cycle or pointer arithmetic.

It's similar in purpose to other things named "magic" elsewhere in the code base, e.g. `ConfigSection::check_magic` and `class Magic`.

PR #678, opened 2025-04-11
RONDB-888: Since UNDO log pages can be completely zero, we need to se…

mronstro:
…t OM_ZEROS_ARE_SPARSE also for UNDO log files when opening

### DIFF
--- a/storage/ndb/include/ndbapi/NdbOperation.hpp
+++ b/storage/ndb/include/ndbapi/NdbOperation.hpp
@@ -1131,7 +1131,9 @@ class NdbOperation {
       OO_TTL_IGNORE    = 0x8000,
       OO_INTERPRETED_INSERT = 0x10000,
       OO_DIRTY_FLAG = 0x20000,
-      OO_TTL_ONLY_EXPIRED = 0x40000
+      OO_TTL_ONLY_EXPIRED = 0x40000,
+      OO_BATCH_SAFE_FLAG = 0x80000,
+      OO_BATCH_UNSAFE_FLAG = 0x100000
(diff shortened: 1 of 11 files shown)


storage/ndb/include/ndbapi/NdbOperation.hpp
```diff
@@ -1131,7 +1131,9 @@ class NdbOperation {
       OO_TTL_IGNORE    = 0x8000,
       OO_INTERPRETED_INSERT = 0x10000,
       OO_DIRTY_FLAG = 0x20000,
-      OO_TTL_ONLY_EXPIRED = 0x40000
+      OO_TTL_ONLY_EXPIRED = 0x40000,
+      OO_BATCH_SAFE_FLAG = 0x80000,
+      OO_BATCH_UNSAFE_FLAG = 0x100000
```
svenssonaxel, line 1136: Why do you need two flags? Shouldn't we assume batching is unsafe if safe flag is 0?

svenssonaxel (changes requested):

storage/ndb/include/ndbapi/NdbOperation.hpp
mronstro: Nope, safe gives us the ability to use batching when deemed unsafe, unsafe means it is unsafe to batch even when considered safe

mronstro: Batching or not is controlled by several variables and conditions

svenssonaxel (approved):

PR #680, opened 2025-04-22
FSTORE-1729: Failed to get project id when project name contains upper case

smkniazi:


storage/ndb/rest-server/data-access-rondb/src/feature_store/feature_store.h
```diff
@@ -65,9 +65,9 @@ typedef struct Serving_Key {
 
 /**
  * Find project ID using the feature store name
- * SELECT id AS project_id FROM project WHERE projectname = feature_store_name
+ * SELECT roject_id FROM feature_store WHERE id = feature_store_id
```
kennethmhc:
```suggestion
 * SELECT project_id FROM feature_store WHERE id = feature_store_id
```

kennethmhc (dismissed):

kennethmhc (approved):

PR #682, opened 2025-04-25
RONDB-753 TTL FEATURE [PATCH 9]

KernelMaker:
support TTL purge window

### DIFF
--- a/storage/ndb/rest-server2/server/src/ttl_purge.cpp
+++ b/storage/ndb/rest-server2/server/src/ttl_purge.cpp
@@ -796,7 +796,8 @@ bool TTLPurger::UpdateLocalCache(const std::string& db,
       } else {
         // check mysql.ttl_purge_nodes
         // TODO(zhao): handle ttl_purge_tables as well
-        if (db == kSystemDBName && table == kTTLPurgeNodesTabName) {
+        if (db == kSystemDBName &&
+           (table == kTTLPurgeNodesTabName || table == kTTLPurgeCtrlTabName)) {
           updated = true;
         }
       }
@@ -812,7 +813,8 @@ bool TTLPurger::UpdateLocalCache(const std::string& db,
     } else {
       // check mysql.ttl_purge_nodes
       // TODO(zhao): handle ttl_purge_tables as well
-      if (db == kSystemDBName && table == kTTLPurgeNodesTabName) {
+      if (db == kSystemDBName &&
+           (table == kTTLPurgeNodesTabName || table == kTTLPurgeCtrlTabName)) {
         updated = true;
       }
     }
@@ -994,6 +996,7 @@ void TTLPurger::PurgeWorkerJob() {
   Int64 packed_now = 0;
   NdbRecAttr* rec_attr[3];
   bool use_index = false;
+  Uint32 purge_window = 0;
 
   g_eventLogger->info("[TTL PWorker] Started");
   purged_pos_.clear();
@@ -1067,6 +1070,14 @@ void TTLPurger::PurgeWorkerJob() {
       g_eventLogger->info("Not the configured purging node, skip purging...");
       sleep(2);
       continue;
+    }
+
+    if (GetPurgeWindow(&purge_window, update_objects) == false) {
+      g_eventLogger->info("[TTL PWorker] Failed to get purge window, "
+                          "error: %u(%s). Retry...",
+                          watcher_ndb_->getNdbError().code,
+                          watcher_ndb_->getNdbError().message);
+      goto err;
     }
 
     GetNow(encoded_now, false);
@@ -1227,6 +1238,7 @@ retry_trx:
           goto err;
         }
         index_scan_op->setPartitionId(iter->second.part_id);
+        index_scan_op->setTTLPurgeWindowSize(purge_window);
         /* Index Scan */
         Uint32 scanFlags =
          /*NdbScanOperation::SF_OrderBy |
@@ -1468,6 +1480,7 @@ retry_trx:
           goto err;
         }
         scan_op->setPartitionId(iter->second.part_id);
+        scan_op->setTTLPurgeWindowSize(purge_window);
         Uint32 scanFlags = NdbScanOperation::SF_OnlyExpiredScan;
         if (scan_op->readTuples(NdbOperation::LM_Exclusive, scanFlags,
                                 1, iter->second.batch_size) != 0) {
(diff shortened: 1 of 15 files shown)


storage/ndb/src/kernel/blocks/dbtc/DbtcMain.cpp
```diff
@@ -15843,6 +15843,8 @@ void Dbtc::execSCAN_TABREQ(Signal *signal) {
   scanptr.p->m_scan_dist_key = scanTabReq->distributionKey;
   scanptr.p->m_scan_dist_key_flag = ScanTabReq::getDistributionKeyFlag(ri);
 
+  scanptr.p->m_ttl_purge_window_size = scanTabReq->ttlPurgeWindowSize;
```
mronstro: I don't see any upgrade code here, thus if an old NDB API sends a SCAN_TABREQ the value of ttl_purge_window_size is undefined which from what I can see will cause issues.

storage/ndb/src/kernel/blocks/dbtc/DbtcMain.cpp
```diff
@@ -18102,6 +18104,7 @@ bool Dbtc::sendScanFragReq(Signal *signal, ScanRecordPtr scanptr,
   req->clientOpPtr = scanFragP.p->m_apiPtr;
   req->batch_size_rows = scanP->batch_size_rows;
   req->batch_size_bytes = scanP->batch_byte_size;
+  req->ttl_purge_window_size = scanP->m_ttl_purge_window_size;
```
mronstro: An alternative approach is to add an extra bit and only set this variable when it is actually used, this approach adds 32 bits of overhead for each SCAN_TABREQ which can affect performance a few percentage points in some cases.

storage/ndb/src/kernel/blocks/dbtup/DbtupExecQuery.cpp
```diff
@@ -2124,6 +2126,7 @@ bool Dbtup::execTUPKEYREQ(Signal* signal,
       ndbrequire(false);
       regOperPtr->ttl_ignore = lqhOpPtrP->ttl_ignore;
     }
+    req_struct.ttl_purge_window_size = lqhScanPtrP->m_ttl_purge_window_size;
```
mronstro, line 2129: No checks that the value is well defined, neither here nor in DBLQH nor in DBTC.

storage/ndb/src/kernel/blocks/dbtup/DbtupExecQuery.cpp
```diff
@@ -2165,6 +2168,12 @@ bool Dbtup::execTUPKEYREQ(Signal* signal,
     }
 #endif  // TTL_DEBUG
   }
+  if (req_struct.ttl_purge_window_size != 0 && !regOperPtr->ttl_only_expired) {
```
mronstro, line 2171: Gather this is the upgrade code, if so there should not be any printout here since this could essentially be printed on each signal, millions of times per second.

storage/ndb/src/ndbapi/NdbOperation.cpp
```diff
@@ -164,6 +164,7 @@ int NdbOperation::init(const NdbTableImpl *tab, NdbTransaction *myConnection) {
   theInterpretIndicator = 0;
   theInterpretInsertIndicator = 0;
   theDistrKeyIndicator_ = 0;
+  theTTLPurgeWindowSize_ = 0;
```
mronstro: Gather the extra _ in the name can be removed, don't think it serves any purpose

storage/ndb/rest-server2/server/src/ttl_purge.cpp
```diff
@@ -1844,6 +1857,138 @@ bool TTLPurger::GetShard(Int32* shard, Int32* n_purge_nodes,
   return false;
 }
 
+bool TTLPurger::GetPurgeWindow(Uint32* purge_window, bool update_objects) {
```
mronstro, line 1860: From what I understand of this code we will do a PK lookup for each scan to get the TTL purge window from a table. That sounds like a lot of overhead, have I understood this correctly?

storage/ndb/src/kernel/blocks/dbtc/DbtcMain.cpp
KernelMaker: Here’s what I was planning to discuss with you after you get back.
The main issue is that ScanTabReq now has an optional field distributionKey added at the end, which makes compatibility handling a bit tricky — not just when a new version receives packets from an old version, but also the other way around, since the old version doesn’t have any logic for handling packet versions.

storage/ndb/src/kernel/blocks/dbtc/DbtcMain.cpp
KernelMaker: I see your point. Personally, I think adding one more field keeps things clearer and simpler, and 32 bits shouldn’t be a big deal performance-wise. But happy to go with your suggestion if we spot any regressions during benchmarking.

storage/ndb/src/kernel/blocks/dbtup/DbtupExecQuery.cpp
KernelMaker: m_ttl_purge_window_size is initialized to 0 in the constructor, and then assigned the value from scanFragReq->ttl_purge_window_size.
Not quite sure what you meant here — were you referring to a compatibility check?

storage/ndb/src/kernel/blocks/dbtup/DbtupExecQuery.cpp
KernelMaker: It’s not meant to handle upgrades.
The purpose of this code is to catch any unexpected requests and surface them.
Here, I want to ensure that ttl_purge_window_size is only set when ttl_only_expired is true — otherwise, it would mean some other unexpected part of the code is incorrectly constructing the scan request.
If everything works as expected, this if statement should never be triggered.
We can remove it after one minor version cycle.

storage/ndb/src/ndbapi/NdbOperation.cpp
KernelMaker: Okay. I just followed the previous code here, 'theDistrKeyIndicator_'

storage/ndb/rest-server2/server/src/ttl_purge.cpp
KernelMaker: Yes and no.
There’s one PK lookup per purge iteration (where we purge all TTL tables with a limited batch size), not per scan.
The frequency isn’t high — it’s about one PK lookup every 2 seconds. I believe RonDB can easily handle that :)
The main reason I’m doing this is to ensure that any change to the purge window is detected and applied as quickly as possible, since it directly impacts the data.

KernelMaker:
Update: Maintain compatibility in ScanTabReq and ScanFragReq.

mronstro (approved):

PR #684, opened 2025-04-30
[RONDB-892] Fix file system password handling bugs

svenssonaxel:
- Do not use uninitialized memory to generate encryption key, as that could lead to unrecoverable data.
- Check that password is not too long, as that would lead to memory corruption.

### DIFF
--- a/storage/ndb/src/kernel/ndbd.cpp
+++ b/storage/ndb/src/kernel/ndbd.cpp
@@ -1156,6 +1156,12 @@ ndbd_run(bool foreground, int report_fd,
           "empty password not allowed");
       ndbd_exit(-1);
     }
+    if (pwd_size > MAX_BACKUP_ENCRYPTION_PASSWORD_LENGTH) {
+      g_eventLogger->info(
+          "Invalid filesystem password, "
+          "too long");
+      ndbd_exit(-1);
+    }
 
     memcpy(globalData.filesystemPassword, pwd, pwd_size);
     globalData.filesystemPassword[pwd_size] = '\0';


storage/ndb/src/kernel/blocks/ndbcntr/NdbcntrMain.cpp
```diff
@@ -6187,6 +6187,8 @@ void Ndbcntr::open_secretsfile(Signal *signal,
   ekm.length = globalData.filesystemPasswordLength;
   memcpy(&ekm.data, globalData.filesystemPassword, globalData.filesystemPasswordLength);
   ekm.data[globalData.filesystemPasswordLength] = 0;
+  ekm.data[globalData.filesystemPasswordLength + 1] = 0;
+  ekm.data[globalData.filesystemPasswordLength + 2] = 0;
```
mronstro: Why is this helping, some comment on why this fix is required would be helpful.

storage/ndb/src/kernel/ndbd.cpp
```diff
@@ -1156,6 +1156,12 @@ ndbd_run(bool foreground, int report_fd,
           "empty password not allowed");
       ndbd_exit(-1);
     }
+    if (pwd_size > MAX_BACKUP_ENCRYPTION_PASSWORD_LENGTH) {
```
mronstro, line 1159: Looks good

storage/ndb/src/kernel/blocks/ndbcntr/NdbcntrMain.cpp
svenssonaxel: We send a number of `Uint32`s in the signal, but we only initialize `length+1` bytes, so some bytes can remain uninitialized. However, it seems that the receiver only uses `length` bytes so perhaps this isn't needed after all.

mronstro (approved):

PR #691, opened 2025-05-17
RONDB-899: Improve performance of data nodes with more than 32 CPUs

mronstro:
Experiments on data nodes with 64 CPUs shows that using 6 send threads was a bit better than using 6 CPUs given that it decreases the mutex bottlenecks on send thread mutex.

Additionally assisting the send threads from LDM and TC threads doesn't provide any benefits, it increases the mutex bottleneck and it sends packets too often hurting also other nodes in the cluster.


storage/ndb/src/common/mgmcommon/thr_config.cpp
```diff
@@ -438,12 +441,35 @@ THRConfig::compute_automatic_thread_config(
...
     Uint32 rem_threads = cpu_cnt -
       (ldm_threads + 2 + tc_threads + recv_threads + send_threads);
-    ldm_threads += rem_threads;
+    /* Handle integer division issues and prioritise more send threads */
+    Uint32 loop_count = 0;
+    while (1) {
+      if (rem_threads == 0) break;
+      rem_threads--;
+      send_threads++;
+      if (rem_threads == 0) break;
```
KernelMaker, line 467: It looks almost identical to the code above and below.
The main operations here are a series of -- and ++, so it would be helpful to either make the logic more compact or add some comments explaining what’s going on.

KernelMaker (approved):

PR #696, opened 2025-05-22
RONDB-905: Hint TC placement for REST API server

mronstro:
Previously no hinting of TC placement was done in REST API Server. The REST API server uses NdbRecord variant of NDB API and there is a simple way of using NdbRecord and pointer to the row for startTransaction call. However one needed to move the setting of the row from setup_read_operation to a new method setup_primary_keys. Renamed setup_transaction and setup_read_operation to setup_transactions and setup_read_operations. Fixed problem introduced by previous patch when setting BatchMaxSize, the change needed to be in init_defaults.go as well for MTR tests to pass.

The special case where all operations in the batch were invalid was handled in a few places.

### DIFF
--- a/storage/ndb/rest-server2/server/test_go/internal/config/init_defaults.go
+++ b/storage/ndb/rest-server2/server/test_go/internal/config/init_defaults.go
@@ -50,7 +50,7 @@ func newWithDefaults() AllConfigs {
 			BufferSize:          5 * 1024 * 1024,
 			GOMAXPROCS:          -1,
 			PreAllocatedBuffers: 32,
-			BatchMaxSize:        256,
+			BatchMaxSize:        1024,
 			OperationIDMaxSize:  256,
 		},
 		GRPC: GRPC{
--- a/storage/ndb/rest-server2/server/src/config_structs_def.hpp
+++ b/storage/ndb/rest-server2/server/src/config_structs_def.hpp
@@ -18,6 +18,11 @@
  */
 
 /*
+ * NOTE:
+ * -----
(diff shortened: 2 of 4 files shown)


storage/ndb/rest-server2/server/test_go/internal/config/init_defaults.go
```diff
@@ -50,7 +50,7 @@ func newWithDefaults() AllConfigs {
 			BufferSize:          5 * 1024 * 1024,
 			GOMAXPROCS:          -1,
 			PreAllocatedBuffers: 32,
-			BatchMaxSize:        256,
```
svenssonaxel, line 53: This isn't the only discrepancy. I see that `RonDB.NodeIDs` and `RonDBMetadataCluster.NodeIDs` also differ in defaults. `RonDBMetadataCluster.Mgmds` has different default logic (go says it'll copy default from RonDB). `RonDBMetadataCluster.ConnectionPoolSize` has different default, as well as `Testing.MySQL.Servers[].Port` and `Testing.MySQLMetadataCluster.Servers`.

svenssonaxel (changes requested):

storage/ndb/rest-server2/server/test_go/internal/config/init_defaults.go
mronstro: Ok, can be fixed in other PR

svenssonaxel: Opened RONDB-908 for that.

svenssonaxel (approved):

PR #704, opened 2025-05-29
RONDB-899: Improve scaling of large data nodes

mronstro:
As nodes goes larger the receive threads have to split sending to many more tc threads. This includes a lot of wakeups and synchronization and mutex handling.

In this patch we provide a configuration variable that decides whether we should only distribute within the Round Robin Group of the receive thread. This has two advantages, first a receive thread will no longer need to send signals to a great number of tc threads, instead it will focus on sending to 2-4 tc threads normally. We also introduce a variable that sets the optimal round robin group size. Both of these new configuration variables enables us to experiment in benchmarks the efficiency of this approach.

The second benefit comes from the fact that the Round Robin Group shares normally at least a L3 CPU cache. This means that communication to the selected tc threads should be more efficient as well.

There is likely still some scaling issues remaining since there is a large number of LDM threads still for the tc threads to communicate with.

### DIFF
--- a/storage/ndb/include/kernel/ndb_limits.h
+++ b/storage/ndb/include/kernel/ndb_limits.h
@@ -1,6 +1,6 @@
 /*
    Copyright (c) 2003, 2024, Oracle and/or its affiliates.
-   Copyright (c) 2021, 2024, Hopsworks and/or its affiliates.
+   Copyright (c) 2021, 2025, Hopsworks and/or its affiliates.
 
    This program is free software; you can redistribute it and/or modify
    it under the terms of the GNU General Public License, version 2.0,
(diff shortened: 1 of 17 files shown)


KernelMaker:
do_parse_auto seems to have changed more than just introducing some parameters.
I suggest that these newly introduced parameters should be disabled by default. If they’re disabled, the rest of the behavior should remain unchanged to avoid introducing any regression (which is especially important right now since the on-call shift is active).
For now, these new parameters should only be enabled for benchmarking. Once they’re fully tested (at least in benchmarks), we can consider enabling them by default in a future iteration. This is just my personal suggestion on this patch, not some strong comment which must to listen.

⸻

One other point, though not directly related to this patch:
To reduce the risk of potential regressions, we should make sure that patches only include the minimal changes needed for their original purpose. Let’s avoid piggybacking unrelated changes in the same patch. In my experience, this is crucial for avoiding accidental regressions and also for keeping the git history clean and easy to trace. This makes it easier to quickly roll back a commit in an emergency — something that’s becoming increasingly important with the on-call responsibilities.

mronstro:
The default of the configuration parameters is the same as the behaviour before, so the default behaviour have not changed in any sense.

KernelMaker (approved):

PR #717, opened 2025-06-13
RONDB-787: Decrease minimum size of RedoBuffer

mronstro:


KernelMaker:
If I understand correctly, this patch aims to reduce memory usage by shrinking the redo log buffer size by half?

I do have some concerns about the potential impact on write performance — could this introduce any regressions under heavy load?
It might be worth running a performance comparison to evaluate the trade-offs more thoroughly.

Also, would it be possible to make this configurable rather than hard-coded? That way, we could tune it very quickly (more flexibly) in production environments if needed

KernelMaker (approved):

PR #744, opened 2025-07-31
RONDB-939: Fixed malloc error checks

smkniazi:


storage/ndb/rest-server/data-access-rondb/src/feature_store/feature_store.cpp
```diff
@@ -514,6 +514,12 @@ RS_Status find_training_dataset_join_data_int(Ndb *ndb_object, int feature_view_
   // freed by CGO
   *tdjs_size = tdjsv.size();
   void *ptr  = (Training_Dataset_Join *)malloc(tdjsv.size() * sizeof(Training_Dataset_Join));
+  if (ptr == nullptr) {
+    return RS_SERVER_ERROR(ERROR_038);
+  }
+  if (ptr == nullptr) {
```
svenssonaxel: Duplicate code

svenssonaxel (changes requested):
Also add checks for `malloc` calls in `rdrs_rondb_connection.cpp`.

smkniazi (commented):
Thanks, removed the duplicated code

svenssonaxel:
@smkniazi and what about `rdrs_rondb_connection.cpp`?

smkniazi:
I did not feel the need to change it as it is a constructor. if malloc fails, then RDRS will not start

svenssonaxel:
> I did not feel the need to change it as it is a constructor. if malloc fails, then RDRS will not start

Sure, but exiting due to a segfault isn't very graceful.

smkniazi (commented):
fixes for rondb_connections

svenssonaxel (approved):

PR #748, opened 2025-08-13
RONDB-932: Ensure releasing records in correct order to avoid long wh…

mronstro:
…ile loops

Fix issues introduced in 24.10 where locked rows are not always properly released. Ensure we don't do any extra loops in removeAccLockOp. Optimisation for ReadCommitted scans.

### DIFF
--- a/storage/ndb/src/kernel/blocks/dblqh/DblqhMain.cpp
+++ b/storage/ndb/src/kernel/blocks/dblqh/DblqhMain.cpp
@@ -17003,9 +17003,21 @@ void Dblqh::scanReleaseLocksLab(Signal *signal,
   Fragrecord::FragStatus fragstatus = fragptr.p->fragStatus;
   ndbrequire(is_scan_ok(scanPtr, fragstatus));
   check_send_scan_hb_rep(signal, scanPtr, regTcPtr);
+  if (scanPtr->readCommitted) {
+    jamDebug();
+    Uint32 release_counter = scanPtr->scanReleaseCounter;
+    Uint32 curr_rows = scanPtr->m_curr_batch_size_rows;
+    Uint32 num_releases = (curr_rows - release_counter) + 1;
+    if (release_counter >= curr_rows) {
+      num_releases = 1;
+    }
+    scanPtr->scanReleaseCounter += (num_releases - 1);
+    scanLockReleasedLab(signal, regTcPtr);
+    return;
+  }
   while (true) {
-    const Uint32 accOpPtr = scanPtr->readCommitted ?
-      get_acc_ptr_from_scan_record(scanPtr, 0, false) :
+    jamDebug();
+    const Uint32 accOpPtr =
       get_acc_ptr_from_scan_record(scanPtr,
                                    scanPtr->scanReleaseCounter-1,
                                    false);
@@ -18989,8 +19001,7 @@ void Dblqh::scanTupkeyConfLab(Signal* signal,
     } else {
       jam();
       scanPtr->scanReleaseCounter = rows + 1;
-      check_send_scan_hb_rep(signal, scanPtr, regTcPtr);
-      scanLockReleasedLab(signal, regTcPtr);
+      scanReleaseLocksLab(signal, regTcPtr);
       return;
     }
   }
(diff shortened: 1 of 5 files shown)


storage/ndb/src/kernel/blocks/dblqh/DblqhMain.cpp
```diff
@@ -17003,9 +17003,21 @@ void Dblqh::scanReleaseLocksLab(Signal *signal,
...
+  if (scanPtr->readCommitted) {
+    jamDebug();
+    Uint32 release_counter = scanPtr->scanReleaseCounter;
+    Uint32 curr_rows = scanPtr->m_curr_batch_size_rows;
+    Uint32 num_releases = (curr_rows - release_counter) + 1;
+    if (release_counter >= curr_rows) {
+      num_releases = 1;
+    }
+    scanPtr->scanReleaseCounter += (num_releases - 1);
+    scanLockReleasedLab(signal, regTcPtr);
```
KernelMaker, line 17015: What would be the difference if we directly set
scanPtr->scanReleaseCounter = scanPtr->m_curr_batch_size_rows
and call scanLockReleasedLab(signal, regTcPtr) inside this if block?
According to the code at the end of Dblqh::scanReleaseLocksLab(), they should be equivalent before calling scanLockReleasedLab().

mronstro: There are two cases where scanReleaseCounter = rows + 1 and m_curr_batch_rows = rows
In this case we will add 0 to scanReleaseCounter and thus not have scanReleaseCounter ==
m_curr_batch_size_rows and this changes the code path in scanLockReleasedLab and
I notice that for aggregate queries you even have a printout if this code is reached.

KernelMaker: Can you show me where the cases "scanReleaseCounter = rows + 1 and m_curr_batch_rows = rows" are? I didn't find it.

mronstro: Here is the code:
  // 'time_passed' is in slices of 10ms
  const Uint32 time_passed = cLqhTimeOutCount - tcConnectptr.p->tcTimer;
  if (unlikely(rows && time_passed > 1) &&
      (refToMain(scanPtr->scanApiBlockref) != DBSPJ || time_passed > 10 ) &&
      (!scanPtr->m_aggregation || scanPtr->m_agg_n_res_recs == 0)) {
    /* PA related
     * explains up if
     * [PA-COMMENT]
     * In pushdown aggregation mode, if we are using filter, then time_pased &&
     * rows > 0 could be true, but it doesn't represent that there are no
     * remaining  aggregation results left in the aggregation_interpreter since
     * the agg batch size limitation hasn't been reached and the results are
     * cached in interpreter temporarily, if we don't skip here, it would send
     * a scanfragconf to TC and the completed ops in scanfragconf is 0, it
     * seems that if the TC received all scanfragconf with ops=0 from all
     * fragments, TC would regard the scan process is completed(but actually
     * it isn't in this situation) and tell completed status to API, API would
     * stop without waiting for the remaining results, which could cause
     * incorrect error in the final results.
     *
     * We can simply just check scanPtr->m_aggregation here to make aggregation
     * scan skip this if. But I would like to introduce an extra one
     * scanPtr->m_agg_n_res_recs to achieve this,
     * which can help us trace and show more details of aggregation scan process.
     */

    /* -----------------------------------------------------------------------
     *  WE NEED TO ENSURE THAT WE DO NOT SEARCH FOR THE NEXT TUPLE FOR A
     *  LONG TIME WHILE WE KEEP A LOCK ON A FOUND TUPLE. WE RATHER REPORT
     *  THE FOUND TUPLE IF FOUND TUPLES ARE RARE. If more than 1 ms passed we
     *  send the found tuples to the API. For requests coming from SPJ we allow
     *  scans to go on for an extended period of 10ms
     * -----------------------------------------------------------------------
     */
    scanPtr->scanReleaseCounter = rows + 1;
    scanReleaseLocksLab(signal, tcConnectptr.p);
    return;

KernelMaker: Now I see it, thanks.
I’m just curious — why set scanPtr->scanReleaseCounter = rows + 1 to make it larger than m_curr_batch_rows here?

KernelMaker (approved):

PR #763, opened 2025-09-10
RONDB-956: More test cases and avoiding to release sections before used

mronstro:


storage/ndb/src/kernel/blocks/dbdict/Dbdict.cpp
```diff
@@ -139,7 +139,7 @@
 //#define DO_TRANSIENT_POOL_STAT 1
 //#define DEBUG_HASH 1
 //#define DEBUG_QUOTAS_EXTRA 1
-//#define DEBUG_QUOTAS 1
+#define DEBUG_QUOTAS 1
```
KernelMaker, line 142: Is this change intentional?

KernelMaker (approved):

storage/ndb/src/kernel/blocks/dbdict/Dbdict.cpp
mronstro: Nope, but I can remove it in separate patch since it won't affect the release build

PR #779, opened 2025-11-07
More than 255 node ids

mronstro:


storage/ndb/src/kernel/blocks/qmgr/QmgrMain.cpp
```diff
@@ -3153,12 +3171,13 @@ void Qmgr::initData(Signal *signal) {
 
   Uint32 sum = 0;
   ArbitSignalData *const sd = (ArbitSignalData *)&signal->theData[0];
+  NodeBitmaskPOD arbit_mask;
   for (unsigned rank = 1; rank <= 2; rank++) {
     sd->sender = getOwnNodeId();
     sd->code = rank;
     sd->node = 0;
     sd->ticket.clear();
-    sd->mask.clear();
```
KernelMaker, line 3161: sd->mask is still defined in the ArbitSignalData, but it seems not to be used anymore

KernelMaker (approved):

PR #789, opened 2025-12-10
RONDB-979: Port RONDB-963 to RDRS2 for Spine Features Support

smkniazi:


storage/ndb/rest-server2/server/src/batch_feature_store_ctrl.cpp
```diff
@@ -461,3 +480,20 @@ void fillPassedFeaturesMultipleEntries(
     }
   }
 }
+
+void fixSpineFGStatus(
+    std::vector<std::vector<std::vector<char>>> &features,
+    const std::vector<std::unordered_map<std::string, std::vector<char>>> &passedFeatures,
```
mronstro: Both passedFeatures and indexLookup looks like they are not used in the function, so why pass them?

smkniazi (commented):
removed unused code

mronstro (approved):

PR #792, opened 2025-12-11
Fix merge issue in Suma.cpp, added test case, fix with FetchAllowed=2

mronstro:


storage/ndb/ndbapi-examples/ndbapi_m/ndbapi_m.cpp
```diff
@@ -0,0 +1,568 @@
+/*
```
KernelMaker, line 1: It looks like the example file I sent you last week for scan debugging. Was it accidentally included in this patch?

PR #808, opened 2026-01-20
RONDB-991: The RonDB REST API pk-read endpoint was returning null values for columns defined with NOT NULL constraints.

smkniazi:
For NOT NULL columns, NDB's default record can have null bit storage
  that overlaps with actual data storage. The code was checking the null bits without
  verifying if the column was actually nullable, causing it to misinterpret data bytes
  as null indicators.  Added a check for col->getNullable()to only check null bits for
  columns that are actually nullable.

### DIFF
--- a/storage/ndb/rest-server2/server/src/db_operations/pk/pkr_operation.cpp
+++ b/storage/ndb/rest-server2/server/src/db_operations/pk/pkr_operation.cpp
@@ -640,7 +640,8 @@ RS_Status KeyOperation::write_col_to_resp(Uint32 colIdx,
     Uint32 null_bit_in_byte;
     bool null_value = NdbDictionary::getNullBitOffset(
       ndb_record, col_id, null_byte_offset, null_bit_in_byte);
-    if (null_value) {
+    // Only check null bits for columns that are actually nullable
+    if (null_value && col->getNullable()) {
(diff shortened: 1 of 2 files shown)


mronstro (dismissed):

storage/ndb/rest-server2/server/src/db_operations/pk/pkr_operation.cpp
```diff
@@ -640,7 +640,8 @@ RS_Status KeyOperation::write_col_to_resp(Uint32 colIdx,
     Uint32 null_bit_in_byte;
     bool null_value = NdbDictionary::getNullBitOffset(
       ndb_record, col_id, null_byte_offset, null_bit_in_byte);
-    if (null_value) {
+    // Only check null bits for columns that are actually nullable
+    if (null_value && col->getNullable()) {
```
svenssonaxel: If `null_value` is only used in this branch, then perhaps `col->getNullable()` should be evaluated before `getNullBitOffset`.

svenssonaxel:
If this fixes RONDB-792, you should also enable the test case in `/rondb/mysql-test/suite/rdrs2/disabled.def`.

svenssonaxel (approved):

PR #825, opened 2026-02-03
RONDB-1003: Add support for TIMESTAMP as primary key column in ClusterJ

smkniazi:
ClusterJ previously did not support TIMESTAMP columns as primary keys, even though the native NDB API supports this. This change adds full support for TIMESTAMP/TIMESTAMP2 primary key columns.


storage/ndb/clusterj/clusterj-core/src/main/java/com/mysql/clusterj/core/metadata/DomainFieldHandlerImpl.java
```diff
@@ -398,6 +400,11 @@ public DomainFieldHandlerImpl(
                     this.objectOperationHandlerDelegate = objectOperationHandlerKeyJavaSqlDate;
                     this.type = java.sql.Date.class;
                     break;
+                case Timestamp:
```
mronstro: Should we really support Timestamp, Timestamp2 replaced Timestamp quite a few years ago?

mronstro: From the patch it seems we only really support Timestamp2, so should not support Timestamp.

mronstro (dismissed):
With removal of Timestamp support, keeping Timestamp2 or reasoning why Timestamp is supported and should be.

smkniazi:
@mronstro removed `timestamp`. please review again

mronstro (approved):

PR #844, opened 2026-02-12
RONDB-1017: Add SET MaxDiskWriteSpeed command to MGM client

mronstro:
Add a proper MGM client command to set MaxDiskWriteSpeed at runtime, replacing the need to use the raw DUMP 100002 command. The new command updates the persistent configuration, the runtime value on data nodes, and the node's local ConfigValues so ndbinfo reports the new value.

Syntax: <id> SET MaxDiskWriteSpeed <value>
        ALL SET MaxDiskWriteSpeed 100M

The implementation uses a semi-generic SetConfigParam signal that carries a config key and Uint64 value, making it easy to extend for additional parameters in the future. Adding a new parameter only requires changes in Cmvmi.cpp (runtime dispatch) and CommandInterpreter.cpp (client-side command).

Includes MTR test (ndb_config_set) and CLAUDE.md developer guide.

### DIFF
--- a/storage/ndb/CLAUDE.md
+++ b/storage/ndb/CLAUDE.md
@@ -1,0 +1,154 @@
+# CLAUDE.md — NDB Storage Engine
+
+## Adding a New Runtime-Settable Configuration Parameter
+
+The `SET` command in the MGM client allows changing configuration parameters on running data nodes. It updates the persistent configuration, the runtime value, and ndbinfo reporting — all in one command.
+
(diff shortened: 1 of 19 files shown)


storage/ndb/CLAUDE.md
```diff
@@ -0,0 +1,154 @@
+# CLAUDE.md — NDB Storage Engine
```
KernelMaker, line 1: Not sure if it’s a good idea to put a single feature description in CLAUDE.md, since it’s a public file that would be read by all the other Claude sessions.

PR #886, opened 2026-03-10
RONDB-1033: Partial availability is ok to start with, health endpoint…

mronstro:
… should report 503 when not ready and fixed a Drogon issue

### DIFF
--- a/storage/ndb/rest-server2/extra/drogon/trantor-1.5.21/trantor/net/inner/Socket.cc
+++ b/storage/ndb/rest-server2/extra/drogon/trantor-1.5.21/trantor/net/inner/Socket.cc
@@ -107,7 +107,14 @@ void Socket::closeWrite()
     if (::shutdown(sockFd_, SD_SEND) < 0)
 #endif
     {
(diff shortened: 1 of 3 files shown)


storage/ndb/rest-server2/extra/drogon/trantor-1.5.21/trantor/net/inner/Socket.cc
```diff
@@ -107,7 +107,14 @@ void Socket::closeWrite()
     if (::shutdown(sockFd_, SD_SEND) < 0)
 #endif
     {
-        LOG_SYSERR << "sockets::shutdownWrite";
+        if (errno == ENOTCONN)
```
KernelMaker, line 110: The fix looks good to me. There’s just one thing we may want to confirm with @smkniazi 
I discussed the fix in Dragon with Mikael yesterday. It’s not the root cause in this case, but Mikael thinks it could be a bug. Since this is a change to third-party code, please double-check it if you have more context.

smkniazi (approved):
One needs to test it

PR #927, opened 2026-05-19
RDMA transporter (RoCE / IB Verbs) for 24.10

elemberger-peak:
Sister PRs: #927 (24.10) · #933 (25.10) · #934 (26.02)

## Summary

Adds an opt-in native RDMA transporter (RoCE / IB Verbs) alongside `TCP_Transporter` / `SHM_Transporter`, plus a 12-phase performance/correctness series, a cluster-wide API↔DB gate (off by default), and the tuning/CI/operator fixes we hit on real hardware. Sharing for feedback against `24.10-main`.

## What's included (24 commits, +5750 / -49, 22 files)

- **`RDMA_Transporter`** — new transporter class with SEND/RECV-based wire protocol, credit-based flow control, and an out-of-band exchange of QP attributes (QPN, PSN, GID/LID) over the existing control socket. Framing: 24-byte header + payload, `MAX_*_MESSAGE_BYTESIZE = 32768`.
- **Config schema** — new `[RDMA]` connection section: `RdmaSendBufferMemory`, `RdmaRecvBufferMemory`, `RdmaQueueDepth`, `RdmaInlineThreshold`, `RdmaCompletionPollBudget`, `RdmaSpintime`, `RdmaDevice`, `RdmaPort`, `RdmaGidIndex`, `RdmaTrafficClass`, `RdmaRetryCount`, `RdmaRnrRetryCount`.
- **Cluster-wide API↔DB gate** (`AllowApiToDbRdma` in `[SYSTEM]`, default `false`) — DB↔DB RDMA is always allowed; API↔DB RDMA requires the explicit cluster-wide opt-in. Field testing showed DB↔DB RDMA improves every workload while API↔DB RDMA only beats TCP under highly concurrent client traffic, so we keep it opt-in by default and let operators flip a single boolean to A/B test it. The validator emits a clear error pointing back at the parameter when API↔DB `[RDMA]` sections appear without the gate. `[SYSTEM]` must precede `[RDMA]` in `config.ini` for the lookup to find it (documented in the parameter description and the error message).
- **Build/feature gating** — `WITH_NDB_RDMA` CMake option and `NDB_RDMA_TRANSPORTER_SUPPORTED` runtime guard. Without the flag, all sources compile to no-ops and TCP/SHM behaviour is bit-for-bit unchanged. With the flag set but a binary built without libibverbs symbols (e.g. mgmd), `[RDMA]` sections are rejected with a clear error message.
- **12-phase performance/correctness series**: P1-2 buffer pool / wire-state cacheline grouping, P3 recv ring, P4 batched doorbells, P5 NUMA-aware completion vectors, P6 MR cache, P7 RDMA WRITE negotiation, P8 remote MR geometry, P9 recv via remote WRITE, P10 write probe, P11 recv-immediate guard, P12 per-WR slot probe.
- **Test plumbing** — `RDMA_Transporter-t` TAP test with a libibverbs kernel-symbol stub so unit tests link without pulling in `Packer<SegmentedSectionArg>`'s dependency on `copy()`. `MgmConfig-t` / `testConfigValues-t` extended with `[RDMA]` scenarios; the gate test covers DB-DB positive, API-DB negative without `[SYSTEM]`, API-DB negative with explicit `false`, API-DB positive with `true`, plus API-API and MGM-involved negatives.
- **Default-tuning fix** — `Rdma{Send,Recv}BufferMemory` defaults raised from `2M` → `4M`. With `2M / RdmaQueueDepth=64 = 32768` the per-slot size equalled `MAX_*_MESSAGE_BYTESIZE` and left zero room for the 24-byte framing header. The runtime per-slot validator therefore rejected the default config and `connect_server_impl()` looped on reconnect. `4M` gives 65536 bytes/slot at QD=64 and headroom up to QD=128. Schema minimums unchanged.
- **Operator documentation** — two new plain-text files in `storage/ndb/docs/`:
  - `rdma-transporter.txt` — operator-facing: build flag, `[RDMA]` parameters, the cluster-wide gate, toggle workflow, troubleshooting.
  - `rdma-wakeup.txt` — reviewer-facing: cast of objects (`m_recv_cq`, `m_recv_comp_channel`, etc.), comp-channel/epoll setup, the wake-up timeline (A–E), the benign arm/fire race window, how `RdmaSpintime` and `SpinMethod` interact, multi-transporter clone fd accounting, and an explicit "what is NOT in this branch" section (no `WRITE_WITH_IMM`, no rkey exchange, no solicited-only notifications — Phase 9 of the experimental series is intentionally out of scope for this PR).

## Verification

- `MgmConfig-t` → `ok 1 - MgmConfig`
- `testConfigValues-t` → exit 0 (RDMA roundtrip)
- `RDMA_Transporter-t` → `ok 1 - RDMA_Transporter` (TAP) with `cmake -DCMAKE_BUILD_TYPE=Release -DWITH_NDB_RDMA=ON -DWITHOUT_SERVER=OFF -DWITH_NDBCLUSTER_STORAGE_ENGINE=ON`
- Live two-node DB↔DB and gated API↔DB tested on Mellanox CX-5 (RoCEv2, `mlx5_0`), including mdtest under sustained load.

## Notes / non-goals

- Purely additive: no existing transporter, signal flow, ABI, ConfigValues layout, or wire format is touched outside the new `[RDMA]` section and the new `CFG_SYS_ALLOW_API_TO_DB_RDMA=4` (next free id in the `CFG_SYS_*` namespace).
- `rnr_retry` is currently pinned to 7 at RTS regardless of the configured value; lower values produced transport-retry-exhausted under mdtest. The user-visible setting is preserved so a future patch can re-enable user control without an ABI change.
- Mainline upgrades / downgrades that don't touch `[RDMA]` are unaffected.
- Sister PRs #933 (25.10) and #934 (26.02) carry the same scope on top of those release lines, but use a **per-link** `AllowApiToDbRdma` (inside `[RDMA]`) plus an `NDB_RDMA_ALLOW_API_TO_DB=kill` env-var kill switch instead of the cluster-wide `[SYSTEM]` gate in this PR. Harmonising the two scopes is a separate decision and not in scope for this PR.

Happy to split into smaller per-phase PRs if maintainers prefer that flow.

### DIFF
--- a/storage/ndb/docs/rdma-wakeup.txt
+++ b/storage/ndb/docs/rdma-wakeup.txt
@@ -1,0 +1,288 @@
+RonDB RDMA transporter -- recv-side wake-up reference
+=====================================================
+
+Audience and scope
+------------------
+This file documents the receive-side wake-up path for the RDMA
+transporter on the 24.10-rdma-port line. It is intended for reviewers
+and operators who need to reason about latency, tail behaviour, or
+spin tuning, and complements the user-facing material in
+storage/ndb/docs/rdma-transporter.txt.
+
+The receive path on this branch is pure ibverbs SEND/RECV. The
+WRITE-with-immediate variant (Phase 9 in the experimental series)
+lives on the 25.10 and 26.02 branches and is NOT part of this build;
+nothing in this file refers to it.
+
+Cast of objects
+---------------
+For every connected RDMA transporter (or clone -- see "Multi-
+transporter clones" below):
+
+    m_send_cq            ibv_cq, no completion channel attached.
+                         Driven purely by poll. doSend() reaps
+                         send completions via reap_send_completions()
+                         on every send pass; no thread ever sleeps on
+                         the send CQ.
+
+    m_recv_cq            ibv_cq, bound at create time to
+                         m_recv_comp_channel. The HCA delivers a
+                         readable event on the channel's fd whenever
+                         a CQE lands AND the CQ is currently armed.
+
+    m_recv_comp_channel  ibv_comp_channel. Owns a file descriptor
+                         set to O_NONBLOCK at allocate time; this is
+                         the fd we register with the recv thread's
+                         epoll set.
+
+    m_qp                 RC QP using m_send_cq for SQ completions
+                         and m_recv_cq for RQ completions.
+
+Setup sequence (per transporter, per connect)
+---------------------------------------------
+allocate_verbs_resources() runs in this order:
+
+    1. ibv_open_device / ibv_query_device / ibv_query_port.
+    2. ibv_alloc_pd.
+    3. ibv_create_cq(send) -- no comp_channel.
+    4. ibv_create_comp_channel(recv) and set the fd O_NONBLOCK.
+       O_NONBLOCK is mandatory: handle_recv_comp_event() drains
+       events in a tight loop and relies on ibv_get_cq_event()
+       returning -1/EAGAIN to break out cleanly.
+    5. ibv_create_cq(recv) bound to the recv comp channel.
+    6. ibv_reg_mr for send + recv staging buffers.
+    7. ibv_create_qp (RC, sq_sig_all = 0).
+    8. allocate_send_slot_state / allocate_recv_slot_state.
+    9. Endpoint exchange over the control socket, then QP transitions
+       INIT -> RTR -> RTS. After RTS, run_endpoint_exchange() calls
+       arm_recv_cq() once: this primes the channel so the first
+       inbound CQE wakes the recv thread immediately.
+
+epoll integration (TransporterReceiveData::epoll_add)
+-----------------------------------------------------
+For TCP/SHM transporters the registry adds the control socket fd to
+epoll with `event.data.u32 = trp_id`. For RDMA transporters we add
+two fds:
+
+    1. The control socket fd, with `data.u32 = trp_id`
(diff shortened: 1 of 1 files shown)


mronstro:
I have looked at the PR and it looks good, there might be a few minor issues I would like to comment on, but mostly it is good. It would be good if you could sign either an Individual Contributor Agreement or a Corporate Contributor Agreement with Hopsworks, we have such an agreement that I can send to you. It is also interesting to discuss to which version of RonDB the contribution should be made. RonDB 24.10 is in maintenance mode, so definitely a possibility, however most of our customers are already working on 26.02 which is based on MySQL 8.4.8 compared to 8.4.4 for 24.10. 26.02 also supports more than 2000 nodes compared to 255 in 24.10. The patch is safe to include in production versions of RonDB where RDMA is disabled. We are working on RonDB 26.04 (will likely be something like 26.10 when it is production ready). In this version I am thinking that RDMA support can also be integrated in the production builds that we do. We will maintain the RDMA support after your contribution and any changes that you propose for RDMA will be taken in after a review if deemed ok, these contributions would be under the same agreement, so agreement is only required for the first contribution as long as the contribution refers to the agreement.

elemberger-peak:
Superseded by #945, which ports the same 24.10-rdma-port content to 26.02-main (the latest stable release) per the agreement to land the first merge against 26.02. Closing in favor of #945.

PR #938, opened 2026-05-26
RONDB-1063: Fix crash in large initial start due to ACTIVATE/DEACTIVATE

mronstro:
ACTIVATE_REQ is sent before starting the data node in rondb-helm. To stop the node we deactivate it. This means that when we start a node while another node has started its start the ACTIVATE_REQ can arrive even before we have had time go through Phase 1.

The code ensures that no one can connect (execCONNECT_REP) before start phase 1 have executed. This phase sets my node phase (state variable) to ZSTARTING. If ACTIVATE_REQ arrived before this we would open communication to the activated node. This led sometimes to the problem that the activated node connected before we came to phase 1.

The fix is fairly straightforward, avoid sending OPEN_COMORD if I haven't reached the phase 1 yet. This means that OPEN_COMORD will be sent from phase 1 start for all data nodes. QMGR expected this call to be the first OPEN_COMORD call. When it wasn't we even executed OPEN_COMORD in the wrong branch. So the fix ensured that the phase 1 OPEN_COMORD was restored as the first call to OPEN_COMORD.

We protect both calls to OPEN_COMORD in this manner although only the ACTIVATE_REQ is the one we can reach in practice.

Added test case for it, it was a bit difficult to reproduce, we had to fix the handling of OPEN_COMORD to reproduce the issue. Also we could not follow the exact mechanism from rondb-helm since MTR fails when we use deactivate. Instead we added a DUMP command to set the node as deactivated.

### DIFF
--- a/storage/ndb/src/ndbapi/NdbAggregator.cpp
+++ b/storage/ndb/src/ndbapi/NdbAggregator.cpp
@@ -193,7 +193,7 @@ Int32 NdbAggregator::ProcessRes(char* buf) {
 
         gb_map_->insert(std::pair<GBHashEntry, GBHashEntry>(
               new_entry, new_aggs));
-        agg_res_ptr = reinterpret_cast<AggResItem*>(agg_rec + agg_res_len);
+        agg_res_ptr = reinterpret_cast<AggResItem*>(agg_rec + gb_cols_len);
       }
       DEB_TRACE();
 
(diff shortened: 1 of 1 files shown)


storage/ndb/src/kernel/blocks/qmgr/QmgrMain.cpp
```diff
@@ -3839,11 +3839,20 @@ void Qmgr::execACTIVATE_REQ(Signal *signal)
...
-      g_eventLogger->info("Open up communication to node %u after activation",
-                          nodePtr.i);
-      signal->theData[0] = 0;
-      signal->theData[1] = nodePtr.i;
-      sendSignal(TRPMAN_REF, GSN_OPEN_COMORD, signal, 2, JBB);
+      NodeRecPtr myNodePtr;
+      myNodePtr.i = getOwnNodeId();
+      ptrAss(myNodePtr, nodeRec);
+      if (myNodePtr.p->phase == ZINIT) {
+        g_eventLogger->info("QMGR: activation defers communication to node %u"
```
KernelMaker, line 3846: What would happen next after skipping the sending signal here?

PR #939, opened 2026-05-27
RONDB-1063: Fix crash in large initial start and issues found during testing it

mronstro:


KernelMaker:
Looks good to me. The only thing worth noting is that the commit message in [2967c7c](https://github.com/logicalclocks/rondb/pull/939/commits/2967c7c85a870c360a0470303bf5162e806f35ff) is incorrect; the root cause of the heap corruption is actually [8969910](https://github.com/logicalclocks/rondb/pull/939/commits/896991024a866ee7b2cc168a881590765488d750).

PR #972, opened 2026-06-25
Fix TTL node-restart copy-fragment failure (error 626 -> 2303)

KernelMaker:
Symptom
-------
During a data node restart, the restarting node is force-shutdown in start phase 5 with error 2303 ("System error, node killed during node restart by other node"). The surviving (live) node logs "copyfrag failed, error: 626" and sends COPY_FRAGREF; DBDIH on the restarting node turns that into SYSTEM_ERROR -> 2303. Only TTL tables that have expired-but-not-yet-purged rows are affected.

Trigger
-------
All of the following must hold:
  - the row is present in the restarting node's local data (LCP/REDO);
  - the row was modified on the live node while the target was down, so its rowGCI is above the target's restore GCI and the copy-fragment scan ships it (the node-restart scan GCI filter would otherwise skip an unchanged row);
  - the row is past its TTL expiry (wall-clock) at copy-fragment time;
  - the expired row is still physically present (not yet purged) on the live node.

Root cause
----------
Copy-fragment is physical replication: it must reproduce the live node's fragment verbatim, including expired-but-unpurged TTL rows. TTL is a logical read/purge policy - expired rows still exist physically and must be copied identically or the replica diverges. The copy-fragment scan marks this intent by setting ScanRecord::m_ttl_ignore = 1 in execCOPY_FRAGREQ.

However, the per-row copy operation built by Dblqh::initCopyTc did not propagate that flag onto TcConnectionrec::ttl_ignore (it stayed 0 from seizeTcrec). packLqhkeyreqLab therefore serialized ttl_ignore = 0 into the LQHKEYREQ, and the restarting node (execLQHKEYREQ via getTTLIgnoreFlag) then re-applied TTL expiry to the copy op:
  - a copy ZINSERT is converted to ZUPDATE by handle_nr_copy when the row already exists locally -> handleUpdateReq -> checkTTL -> 626;
  - a copy delete-by-rowid ZDELETE -> handleDeleteReq -> checkTTL -> 626. Either 626 fails the copy-fragment -> COPY_FRAGREF -> 2303.

Fix
---
In Dblqh::initCopyTc, inherit the copy scan's TTL-ignore intent:

    regTcPtr->ttl_ignore = scanptr.p->m_ttl_ignore;

This covers both copy operation types (the row-copy ZINSERT and the delete-by-rowid ZDELETE) and is carried to the restarting node by the existing packLqhkeyreqLab serialization, so the copy op skips checkTTL. There is no replica divergence: the expired row is copied identically to the live node and is hidden on reads / removed by purge as usual once the node rejoins.

Note: this addresses the copy-fragment path only. The LCP-restore and REDO replay recovery paths are tracked separately for follow-up.

Test
----
Add MTR regression test ndb_ttl.ttl_node_restart_copyfrag: create a fixed-width TTL table, force an LCP, stop a data node, UPDATE the row (bumping rowGCI above the stopped node's restore GCI), wait out the TTL, then restart the node and assert it rejoins. On the unfixed code the restarting node crashes with 2303 and the restart waiter times out; with the fix it rejoins cleanly and the expired row is filtered on read. The test pins the session to UTC because checkTTL compares ttl_col against UTC. Uses NoOfReplicas=2 (one node group) so restarting one node triggers a copy-fragment from the surviving replica.

### DIFF
--- a/storage/ndb/src/kernel/blocks/dbtup/DbtupExecQuery.cpp
+++ b/storage/ndb/src/kernel/blocks/dbtup/DbtupExecQuery.cpp
@@ -3004,8 +3004,13 @@ int Dbtup::checkTTL(Tablerec* regTabPtr,
                       "%u.%u.%u %u:%u:%u",
                       dt.year, dt.month, dt.day,
                       dt.hour, dt.minute, dt.second);
-      Uint32 ttl_sec = regTabPtr->m_ttl_sec;
-      ttl_sec += req_struct->ttl_purge_window_size;
+      // Sum in 64-bit: m_ttl_sec can be as large as RNIL-1 and the purge
+      // window is added on top, so a Uint32 add could wrap and defeat the
+      // date_add_interval() overflow guard below (a wrapped small value would
+      // yield a bogus near-term expiry -> premature purge). date_add_interval()
+      // range-checks the result, so no separate upper-bound clamp is needed.
+      const Uint64 ttl_sec =
+          Uint64(regTabPtr->m_ttl_sec) + req_struct->ttl_purge_window_size;
       bool valid_future_dt = true;
       if (ttl_sec != 0) {
         Interval interval;
@@ -3038,11 +3043,12 @@ int Dbtup::checkTTL(Tablerec* regTabPtr,
          * Compare with TTL
          */
         TTL_RONDB_TRACE(req_struct->fragPtrP->fragTableId,
-                        "Get TTL [%u + (%u) = %u], "
+                        "Get TTL [%u + (%u) = %llu], "
                         "expired time: %u.%u.%u %u:%u:%u, "
                         "current time: %u.%u.%u %u:%u:%u",
                         regTabPtr->m_ttl_sec,
-                        req_struct->ttl_purge_window_size, ttl_sec,
+                        req_struct->ttl_purge_window_size,
+                        (unsigned long long)ttl_sec,
                         dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second,
                         curr_dt.year, curr_dt.month, curr_dt.day, curr_dt.hour,
                         curr_dt.minute, curr_dt.second);
--- a/mysql-test/suite/ndb_ttl/r/ttl_ddl_max_sec.result
+++ b/mysql-test/suite/ndb_ttl/r/ttl_ddl_max_sec.result
@@ -1,0 +1,22 @@
+DROP TABLE IF EXISTS test.sz;
+# TTL = RNIL (4294967040) must be rejected on CREATE (not silently stored
+# as the no-TTL sentinel).
+CREATE TABLE test.sz (
+a INT PRIMARY KEY,
+b TIMESTAMP NULL DEFAULT NULL
+) ENGINE=NDB COMMENT="NDB_TABLE=TTL=4294967040@b";
+ERROR HY000: Table storage engine 'ndbcluster' does not support the create option 'The maximum ttl is 4294967039 seconds'
+# the documented maximum TTL (RNIL-1 = 4294967039) is accepted
+CREATE TABLE test.sz (
+a INT PRIMARY KEY,
+b TIMESTAMP NULL DEFAULT NULL
+) ENGINE=NDB COMMENT="NDB_TABLE=TTL=4294967039@b";
+DROP TABLE test.sz;
+# TTL = RNIL added via ALTER must also be rejected
+CREATE TABLE test.sz (
+a INT PRIMARY KEY,
+b TIMESTAMP NULL DEFAULT NULL
+) ENGINE=NDB;
+ALTER TABLE test.sz COMMENT="NDB_TABLE=TTL=4294967040@b";
+ERROR HY000: Table storage engine 'ndbcluster' does not support the create option 'The maximum ttl is 4294967039 seconds'
+DROP TABLE test.sz;
--- a/mysql-test/suite/ndb_ttl/r/ttl_restore_log_expired.result
+++ b/mysql-test/suite/ndb_ttl/r/ttl_restore_log_expired.result
@@ -1,0 +1,38 @@
+DROP TABLE IF EXISTS test.sz;
+CREATE TABLE test.sz (
+col_a INT PRIMARY KEY,
+col_b TIMESTAMP NULL DEFAULT NULL,
+col_c INT
+) ENGINE=NDB COMMENT="NDB_TABLE=TTL=10@col_b";
+# anchor row 1 (col_b NULL -> never expires, control) + row 2 (expires)
+INSERT INTO test.sz VALUES (1, NULL,   100);
+INSERT INTO test.sz VALUES (2, NOW(),  200);
+# --- stall backup AFTER the fragment scan (error 10040 pauses at the
+#     WAIT_GCP phase): the scan captures row 2's ORIGINAL col_b, and the
+#     refresh UPDATE below then lands ONLY in the backup log (so the
+#     restored snapshot copy is the expiring one, exercising log replay).
+Starting backup
+Backup started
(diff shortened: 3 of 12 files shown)


storage/ndb/src/kernel/blocks/dblqh/DblqhMain.cpp
```diff
@@ -10120,6 +10120,19 @@ void Dblqh::exec_acckeyreq(Signal *signal, TcConnectionrecPtr regTcPtr) {
     /*
      * TTL related. Tell DBACC not to convert a duplicate ZINSERT into an
      * in-place TTL update for LCP-restore-originated operations.
+     *
+     * NOTE: this is deliberately NOT extended to REDO replay
+     * (c_executing_redo_log). An insert that replaced an expired row commits as
+     * ZINSERT_TTL (= ZINSERT | 0x08 = 10), but LqhKeyReq's operation field is
+     * only 3 bits, so on replay it arrives here as a plain ZINSERT. Suppressing
```
mronstro: as a plain UPDATE

storage/ndb/src/kernel/blocks/dblqh/DblqhMain.cpp
```diff
@@ -34284,6 +34284,15 @@ void Dblqh::readLogHeader(LogPageRecordPtr &logPagePtr,
...
         readLogwordExec(logPagePtr, logPartPtrP);
   }  // if
+  /*
+   * ZINSERT_TTL is an internal DBLQH operation which is stored verbatim in
+   * old REDO records, but it does not fit in LqhKeyReq's operation field.
+   * Replay it as the underlying ZINSERT; DBACC will re-derive ZINSERT_TTL if
+   * the key restored by the LCP is still present.
+   */
+  if (tcConnectptr.p->operation == ZINSERT_TTL) {
+    tcConnectptr.p->operation = ZINSERT;
```
mronstro, line 34358: ZINSERT -> ZUPDATE

storage/ndb/src/kernel/blocks/dblqh/DblqhMain.cpp
```diff
@@ -34658,7 +34667,9 @@ void Dblqh::sendLqhTransconf(Signal *signal, LqhTransConf::OperationStatus stat,
   LqhTransConf::setLastReplicaNo(reqInfo, tcConnectptr.p->lastReplicaNo);
   LqhTransConf::setSimpleFlag(reqInfo, tcConnectptr.p->opSimple);
   LqhTransConf::setDirtyFlag(reqInfo, tcConnectptr.p->dirtyOp);
-  LqhTransConf::setOperation(reqInfo, tcConnectptr.p->operation);
+  /* ZINSERT_TTL carries ZINSERT in the protocol's three-bit operation field. */
```
mronstro: ZUPDATE

storage/ndb/src/kernel/blocks/dblqh/DblqhMain.cpp
```diff
@@ -9382,6 +9383,24 @@ void Dblqh::execLQHKEYREQ(Signal *signal) {
...
+   * delete-by-PK + reinsert recovery keeps the row count consistent.
+   */
+  regTcPtr->m_restore_op = (senderBlockNo == getRESTORE()) ? 1 : 0;
+  /*
+   * TTL related (companion to the NoTTLDupConvert insert fix). A restore
+   * delete can target an EXPIRED row: the duplicate-key recovery deletes the
+   * stale copy by PK (restore.cpp), and normal replay issues DELETE BY ROWID.
+   * Without ignoring TTL such a delete returns 626 on the expired row, which
+   * would break the recovery. Restore rebuilds physical state, so its deletes
+   * must be expiry-blind.
```
KernelMaker: Add comment here:
Restore only does INSERT and DELETE.
INSERT:
DELETE:

storage/ndb/src/kernel/blocks/dblqh/DblqhMain.cpp
KernelMaker, line 34358: Add a comment to explain why insert works

storage/ndb/src/kernel/blocks/dbtup/DbtupExecQuery.cpp
```diff
@@ -3474,6 +3474,88 @@ int Dbtup::handleUpdateReq(Signal* signal,
...
+          break;
+        }
+        pos += 1 + dsz;
+      }
+    }
+    /* A well-formed index insert always carries NDB$PK (built by DBTC). */
+    ndbrequire(incomingOwner != nullptr);
+    const bool same_owner =
+        (incomingOwnerWords == existingOwnerWords) &&
+        (memcmp(existingOwner, incomingOwner,
```
KernelMaker: charactset

storage/ndb/src/kernel/blocks/dbtup/DbtupExecQuery.cpp
```diff
@@ -3474,6 +3474,88 @@ int Dbtup::handleUpdateReq(Signal* signal,
...
+   * var-data.
+   */
+  if (operPtrP->op_type == ZINSERT_TTL &&
+      operPtrP->ttl_ignore == 0 &&
+      c_lqh->is_unique_hash_index_table(req_struct->fragPtrP->fragTableId)) {
+    jam();
+    const Uint32 ownerAttrId = regTabPtr->noOfKeyAttr;
+    /* Existing owner: read NDB$PK from the existing-valued tuple (m_tuple_ptr). */
+    Uint32 readReq = (ownerAttrId << 16);
+    Uint32 ownerBuf[MAX_KEY_SIZE_IN_WORDS + 2];
```
KernelMaker: try to use dbtup buffer(member variable) instead

storage/ndb/src/kernel/blocks/dbtup/DbtupExecQuery.cpp
```diff
@@ -3474,6 +3474,88 @@ int Dbtup::handleUpdateReq(Signal* signal,
...
+   * updateAttributes below) with the variable-length metadata initialised --
+   * required because write execution does not run prepare_read(), so reading a
+   * variable NDB$PK from the pre-expand tuple would dereference uninitialised
+   * var-data.
+   */
+  if (operPtrP->op_type == ZINSERT_TTL &&
+      operPtrP->ttl_ignore == 0 &&
+      c_lqh->is_unique_hash_index_table(req_struct->fragPtrP->fragTableId)) {
+    jam();
+    const Uint32 ownerAttrId = regTabPtr->noOfKeyAttr;
```
KernelMaker: Put the logic in a separate function

storage/ndb/src/kernel/blocks/dblqh/DblqhMain.cpp
```diff
@@ -23007,6 +23087,25 @@ void Dblqh::initCopyTc(Signal *signal, Operation_t op,
...
+  /*
+   * TTL related
+   * A copy-fragment op physically replicates the live node's row (including
+   * expired-but-not-yet-purged TTL rows). It must NOT re-apply TTL expiry on the
+   * starting node, otherwise the converted copy ZUPDATE / delete-by-rowid ZDELETE
+   * hits checkTTL -> 626 -> COPY_FRAGREF -> the starting node is shut down in
+   * start phase 5 (error 2303). Inherit the copy scan's TTL-ignore intent (set to
+   * 1 in execCOPY_FRAGREQ); packLqhkeyreqLab serializes it to the starting node.
+   */
+  regTcPtr->ttl_ignore = scanptr.p->m_ttl_ignore;
```
KernelMaker: ttl_ignore should be 1 unconditionally here, right?
Then add Ndbrequire on scanptr.p->m_ttl_ignore and skip the if (regTcPtr->ttl_ignore) below

PR #987, opened 2026-07-07
RONDB-1074: Two-level hash scheme for lower latency index scans

mronstro:


storage/ndb/src/ndbapi/NdbDictionaryImpl.cpp
```diff
@@ -59,6 +59,53 @@
...
+    }
+    if (impl.getFullyReplicated()) {
+      error.code = CreateTableRef::WrongPartitionBalanceFullyReplicated;
+      return -1;
+    }
+    if (impl.m_fragmentType == NdbDictionary::Object::UserDefined) {
+      error.code = CreateTableRef::InvalidPartitionHash;
+      return -1;
+    }
+    if (partition_count_known && (partition_count % fanout) != 0) {
```
KernelMaker: This only checks that the fanout divides the partition count, but the interval scan also needs the fanout to divide the hashmap length. DIH resolves each raw hash as `m_map[hashValue % m_cnt]` with `m_cnt` = 3840 by default (240 with the small DefaultHashMapSize), and nothing checks `m_cnt % fanout`.

Example that passes validation: fanout=7 with 14 partitions and the default 3840-bucket map (`map[i] = i % 14`). Since 7 doesn't divide 3840, a grouped interval can straddle the end of the map: `g % 3840 == 3836` gives buckets {3836..3839, 0, 1, 2} -> fragments {0, 1, 2, 3, 0, 1, 2}. `sendDihGetNodeReq()` appends all seven locations without any dedup, so fragments 0-2 get scanned twice with the same bounds and matching rows come back twice. Roughly (fanout-1)/3840 of base keys land on the wrap, so it's rare and silent. Rows are never missed (placement uses the same raw hashes), it's duplicates only.

This also bites power-of-two fanouts on a 240-bucket map (32 doesn't divide 240), and custom hashmaps installed via `setHashMap()` can break the alignment regardless of length.

Options: validate `hashmap_length % fanout == 0` here and in `Dbdict::handleTabInfoInit` (plus something for non-default maps), or dedup the resolved fragment ids in DBTC before starting the fragment scans. Worth a fanout=7 test either way - the current tests all use fanouts that divide 3840, which is why this never showed up.

storage/ndb/src/ndbapi/NdbQueryBuilder.cpp
```diff
@@ -2029,6 +2032,17 @@ Uint32 NdbQueryIndexScanOperationDefImpl::appendPrunePattern(
    */
   if (getOpNo() == 0) return 0;
 
+  /*
+   * DBSPJ prune patterns currently map one prune key to one fragment.
+   * Partition-hash fanout needs one base-key prune key to scan a raw hash
+   * interval of fanout fragments. Until DBSPJ expands prune keys into that
+   * interval, do not push a prune pattern for fanout tables; the child scan
+   * remains correct by scanning all fragments with its normal index bounds.
```
KernelMaker, line 2040: This bail-out only protects clients built with this patch. An older mysqld/API client opens a fanout table as an ordinary table (`SimpleProperties::unpack` ignores unknown keys), still sees the legacy distribution-key metadata, and will serialize T_CONST_PRUNE / T_PRUNE_PATTERN for a pushed child scan like it always did. DBSPJ has the fanout metadata from TC_SCHVERREQ, but neither prune path looks at it - they go through the legacy `computePartitionHash()` and route the child scan to a single fragment, so rows sitting in the other fanout fragments silently disappear from the join result.

DBTC handles exactly this case kernel-side with error 2203; SPJ has no equivalent. Until Phase 5 does real interval expansion, rejecting T_CONST_PRUNE/T_PRUNE_PATTERN in `parseScanFrag()` when `m_partition_hash_fanout > 1` would turn silent row loss during rolling upgrades into a loud error.

storage/ndb/src/ndbapi/ndberror.cpp
```diff
@@ -694,6 +695,7 @@ ErrorBundle ErrorCodes[] = {
   { 796,  DMEC, SE, "Out of schema transaction memory" },
   { 798,  DMEC, AE, "A disk table must not be specified as no logging" },
   { 799,  HA_WRONG_CREATE_OPTION, SE, "Non default partitioning without partitions" },
+  { 800,  HA_WRONG_CREATE_OPTION, SE, "Invalid partition hash definition" },
```
KernelMaker: 800 is in the range this file reserves for TUP, and DbtupMeta.cpp already sets `terrorCode = 800` in release builds when the table id is outside the configured table-record range. With this entry that config mismatch would now be reported as "Invalid partition hash definition", which will send whoever hits it down the wrong path since both errors show up on schema operations.

There's room in the 12xx block CreateTableRef already uses (1224, 1235, 1236) - probably easiest to move it there.

storage/ndb/plugin/ha_ndbcluster.cc
```diff
@@ -9348,6 +9384,97 @@ static bool parsePartitionBalance(
...
+      primary_key_count)
+    return "PARTITION_HASH key counts must match primary key columns";
+
+  if (partition_hash.fanout > 1 && partition_hash.detail_key_count == 0)
+    return "PARTITION_HASH fanout > 1 requires detail key columns";
+
+  if (fully_replicated)
+    return "PARTITION_HASH is not supported for fully replicated tables";
+
+  if (user_defined_partitioning)
```
KernelMaker, line 9470: This rejects RANGE/LIST but lets `PARTITION BY KEY(...)` through - that path keeps HashMapPartition and just sets distribution keys on the columns, so the combination with PARTITION_HASH passes validation. The fanout code then ignores the declared keys completely (the base key is always the first base_count PK columns in key order), so `PARTITION BY KEY(b)` plus a PARTITION_HASH comment silently partitions by something other than what the user wrote.

The declared d-key flags also still go into DictTabInfo, which is what pre-fanout clients prune on - with a subset key declared, an old client prunes on equality over that single column, which makes the SPJ prune issue above much easier to hit.

I'd reject PARTITION_HASH when explicit partition-key columns are declared (or require them to match the base prefix).

storage/ndb/src/kernel/blocks/dbtc/DbtcMain.cpp
```diff
@@ -3171,6 +3177,16 @@ void Dbtc::hash(Signal *signal, CacheRecord *const regCachePtr) {
...
     rondb_calc_hash(tmp, (const char*)&Tdata32[0], keylen, use_new_hash_function);
+    if (!distKey && tabPtrP->m_partition_hash_fanout > 1)
+    {
+      bool ok = handle_partition_hash(tmp,
+                                      &Tdata32[0],
+                                      keylen,
+                                      regCachePtr->tableref,
+                                      0,
+                                      use_new_hash_function);
+      ndbrequire(ok);
```
KernelMaker, line 3188: This ndbrequire is reachable from a malformed request: `handle_partition_hash()` returns false when the KEYINFO is shorter than the fixed key length (`total_len > srcLen`), and keylen isn't validated against the table's key length anywhere before `hash()` runs - it comes straight from the TCKEYREQ. So on a fixed-key fanout table a buggy API client sending a short key crashes the TC node, where the old code just computed a wrong hash and failed further down with a normal error. The other false conditions look unreachable (the counts are validated by DICT, var-key tables go through `handle_special_hash` which fails gracefully), but this one is driven by the wire. Better to abort the operation with an error code.

PR #994, opened 2026-07-13
mgm: bound the nodeid-allocation retry loop and cap MGM API sessions

elemberger-peak:
## Problem
We hit the same `ndb_mgmd` failure three times in one week on a 32-data-node cluster (RonDB-26.02.4, RHEL 9.8, AWS m8i.12xlarge; single mgmd; ~930 NDB API connections from 32 clients; 1984 `[api]` slots):

1. **Per-connection threads/fds are never released after a mass reconnect.** Captures: 32,204 threads with only ONE established TCP connection remaining; 31,887 threads / ~32,133 fds vs 2,080 live connections and 10.4 GB RSS. In this state the MGM API stops responding — `ndb_mgm -e show` times out and every data node loops forever on `Failed to register dynamic ports, error: 110`, so a degraded cluster can never reassemble. Persists >1.5 h; only an mgmd restart recovers.
2. **`ALLOC_NODE_ID_REQ` retried with no backoff.** The cluster log's own suppression shows `Sending ALLOC_NODE_ID_REQ to node 16 — Repeated 17246 times` just 3 seconds after the first send (~10,800/s), sustained for over an hour per episode.

## Root cause
`MgmtSrvr::alloc_node_id_req()` retries forever: `timeout_ms` is only embedded in the request payload for the data node's benefit, the loop has no deadline, and `ss.waitFor()` is called without a timeout. When data nodes keep replying `ALLOC_NODE_ID_REF` `NotMaster` from a non-master (the one REF path with **no sleep**), the loop spins at network round-trip rate. The `MgmApiSession` thread driving it never returns to its command loop, so it never observes the client hanging up — the session is never reaped, leaking one thread + one fd per abandoned registration attempt. `SocketServer` supports a session cap but `ndb_mgmd` never sets one, so nothing bounds the growth.

Both defects compound: every leaked session is a thread stuck in the unbounded alloc loop, and each data-node retry opens another one.

## Fix (all in `storage/ndb/src/mgmsrv/MgmtSrvr.cpp`)
- `alloc_node_id_req()`: enforce a wall-clock deadline (`timeout_ms`, or 20 s when 0) across the retry loop, and poll `waitFor()` in bounded 1 s slices so the deadline is re-checked even with no signal traffic. On expiry return `NO_CONTACT_WITH_DB_NODES`, which callers already handle (data nodes fall back to the existing local-reservation path, API allocation returns a bounded error) — the session thread returns to its command loop, sees EOF, and is reaped normally.
- Back off repeated `NotMaster` redirects: the first redirect to a known master stays immediate (normal one-hop handshake); an unknown/unconfirmed master or a NotMaster ping-pong now sleeps 100 ms per retry (~10 req/s instead of wire rate).
- Cap concurrent MGM API sessions at 4096 via the existing `SocketServer` `maxSessions` mechanism. At the cap the accept loop defers (sleep+retry) instead of accepting unboundedly, so a reconnect storm degrades to delayed connects rather than a dead mgmd.

## Verification
- `ndb_mgmd` builds cleanly (gcc 13, `-DWITH_NDB=1` minimal config) and runs.
- 25 consecutive `ndb_mgm -e show` sessions against the patched mgmd: thread/fd counts return to baseline (12 threads / 21 fds) — normal session lifecycle unaffected.
- The full wedge needs a large cluster mid `--initial` bring-up holding data nodes in a REF loop; not reproducible in a unit environment. Happy to run a patched build on the affected 32-node cluster and report back.

Related: follows up on the node-id-space robustness work in #988 (error 2341 fix) — this addresses the mgmd-side session/retry behaviour observed at the same fleet scale.

### DIFF
--- a/storage/ndb/src/mgmsrv/MgmtSrvr.cpp
+++ b/storage/ndb/src/mgmsrv/MgmtSrvr.cpp
@@ -227,11 +227,25 @@ static int translateStopRef(Uint32 errCode) {
   return 4999;
 }
 
+/*
+ * Upper bound on concurrent MGM API sessions.  Each accepted connection
+ * gets a dedicated session thread + fd; without a bound, a reconnect
+ * storm against a slow or wedged mgmd grows both without limit (>31k
+ * threads / 10 GB RSS observed in the field).  When the cap is reached
+ * the accept loop simply defers accepting (SocketServer::doRun sleeps
+ * and retries), so clients see a delayed connect rather than a dead
+ * mgmd.  The value is sized generously: the widest supported clusters
+ * use ~2k node ids, and MGM sessions (nodeid allocation, config fetch,
+ * transporter handoff, mgm clients) are short-lived.
+ */
+static constexpr unsigned MAX_MGM_API_SESSIONS = 4096;
+
 MgmtSrvr::MgmtSrvr(const MgmtOpts &opts)
     : m_opts(opts),
       _blockNumber(0),
       _ownNodeId(0),
       m_port(0),
+      m_socket_server(MAX_MGM_API_SESSIONS),
       m_local_config(NULL),
       _ownReference(0),
       m_config_manager(NULL),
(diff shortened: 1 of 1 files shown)


mronstro:
Hi Eyal,
I reviewed your patch and it looks correct, so I merged it into 26.02.
We are in the process of releasing a new 26.02 release, it will be 26.02.8
to avoid conflict
with your local 26.02.7 and this version will have this fix and a lot more
fixes we have been
working on recently (mostly TTL fixes).

Rgrds Mikael Ronström

PR #1007, opened 2026-07-22
mtr: cover SET RdmaLogLevel and fix ndb_config_set result (follow-up to #998)

elemberger-peak:
Follow-up to #998 (merged).

#998 added the `SET RdmaLogLevel` `ndb_mgm` command and made `executeSet()` advertise it in its "Supported parameters" message. The matching `mysql-test` update landed just after #998 merged, so `26.02-main` currently has the code change without it.

This PR:
- **Fixes `ndb_config_set.result`** so the existing `SET MaxDiskWriteSpeed` error cases match the current output (`Supported parameters: MaxDiskWriteSpeed, RdmaLogLevel`). Without this the `ndb_config_set` test fails on `26.02-main`.
- **Adds `SET RdmaLogLevel` cases** to `ndb_config_set.test`: ALL and single-node dispatch, `0..3` range validation, a non-numeric value, and the missing-value usage path.

mtr clusters use TCP transporters (no `[RDMA]` sections), so a syntactically valid `SET RdmaLogLevel` parses, validates the range, dispatches (ALL vs single node), fetches config and reports "No RDMA connections found". The test therefore exercises the MGM client command path and input validation, per the review suggestion, rather than an actual RDMA level change (which needs an RDMA-enabled cluster).

### DIFF
--- a/mysql-test/suite/ndb/r/ndb_config_set.result
+++ b/mysql-test/suite/ndb/r/ndb_config_set.result
@@ -25,10 +25,10 @@ MaxDiskWriteSpeed must be between 1M and 1024G
 Invalid value: '100X'. Use a number with optional K, M, or G suffix (e.g., 100M)
 
 Usage: <id> SET <parameter> <value>
-Supported parameters: MaxDiskWriteSpeed
+Supported parameters: MaxDiskWriteSpeed, RdmaLogLevel
 
 Unknown SET parameter: 'FooBar'
-Supported parameters: MaxDiskWriteSpeed
+Supported parameters: MaxDiskWriteSpeed, RdmaLogLevel
 
 select node_id, config_value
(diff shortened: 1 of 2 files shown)


elemberger-peak:
Heads-up for reviewers: in addition to adding `SET RdmaLogLevel` coverage, this PR fixes a **currently-failing `ndb_config_set` mtr test on `26.02-main`**.

#998 merged the `executeSet()` change that lists `RdmaLogLevel` under "Supported parameters", but the matching `ndb_config_set.result` update landed just after that merge. So on `26.02-main` the existing `SET MaxDiskWriteSpeed` error cases now mismatch the recorded output. Merging this repairs the test in addition to adding the new coverage, so it's worth prioritizing.

PR #1012, opened 2026-07-23
Raise API node id ceiling from 2039 to 8191

elemberger-peak:
## Summary

Raises the API node id ceiling from 2039 to 8191 (`ABS_MAX_NODES` 2040 -> 8192,
all-node bitmask 64 -> 256 words), following the pattern RONDB-914 used for the
255 -> 2039 raise. Commits:

1. **Raise API node id ceiling from 2039 to 8191** - constants, a frozen 64-word
   legacy mask type (`NodeBitmask2K`) so every `*_v1` fixed-length wire format keeps
   its exact old layout, a branch-aware `ndbd_support_8k_api_nodes()` gate
   (cutoff **26.02.9** - note 26.02.8 was already released without this feature, so
   the version is bumped to 26.02.9), two-tier QMGR/mgmd version gates, exclusive-
   bounds fixes (incl. a real OOB read in `MgmtSrvr::getNextNodeId()`), a checked
   bitmask-section decoder, a clean v1-pack refusal (previously any v1 `get config`
   against a >16383-section config **aborted mgmd** via `require()` - remotely
   triggerable), and a `set config` payload bound raise 1 MiB -> 8 MiB (a minimal
   2-data-node 8k-slot config measures 1.43 MB packed+base64).
2. **Schema distribution for SQL nodes beyond the slock capacity** - keeps
   `slock BINARY(255)` unchanged; high-id SQL nodes use the ndb_schema_result
   protocol exclusively: off-by-one + `ndbrequire` crash removed in
   `ack_schema_op`, and schema-dist setup is refused with an actionable error if
   the node's id has no slock bit and `ndb_schema` still has the legacy format.
3. **Tests** - new `ndb.ndb_high_nodeid` MTR test (mysqlds at ids 8191/2040/8000
   vs 2039/63/127, bidirectional DDL distribution), result-file updates, and an
   mtr harness fix (`ronsql_cli` lookup blocked the whole ndb suite on non-RDRS
   builds).
4. **Size ndbmtd trp-id indexed arrays from runtime config, not ABS_MAX_NODES** -
   restores small-cluster memory efficiency (review feedback): the mt.cpp
   send-buffer structures were embedded arrays sized by the compile-time
   MAX_NTRANSPORTERS ceiling, costing ~380 MB resident on every data node
   regardless of the configured node ids. They are now allocated at
   ThreadConfig::init() time, sized from the configured max node id (plus the
   node-group multi-transporter margin), capped at the ceiling. Hot paths are
   unchanged (same `base + trp_id` indexing).
5. **Fail only the test, not the whole run, when rdrs2 is missing** - mtr
   harness fix found while running the full multi-suite coverage: on
   non-RDRS builds any test whose configuration defines rdrs processes
   aborted the entire mysql-test-run invocation; it now fails just that test
   with an actionable comment.
6. **Fix ndbcrunch replication tests: restore binlogging server setup** -
   pre-existing suite breakage (since RONDB-210, unrelated to this feature)
   found by the same coverage runs: commenting out `[mysqld.1.crunch]`
   disabled `ndb-log-bin` everywhere AND made the applier the default
   connection, so the crunch_update* tests spun until the testcase timeout.
   Restored the group (ordering documented), plus test-side replication
   state cleanup so check-testcase passes. All 18 ndbcrunch tests pass;
   recorded .result files unchanged.
7. **Speed up mgmd node id allocation under mass API connects** - two
   mgmd-local, protocol-compatible changes addressing the fleet cold-start
   bottleneck measured in the storm tests below: adaptive exponential
   backoff (2 ms first retry, jitter, 65-128 ms cap) instead of a fixed
   100 ms sleep on `AllocNodeIdRef::Busy`, and a rotated wildcard
   candidate-scan start (own-node-id hash + per-request round robin) so
   multiple mgmds stop racing for the same lowest free id. Rotation only
   applies to API allocations with >= 64 equally preferred candidates -
   small configurations keep the traditional deterministic assignment
   (existing MTR results unchanged).
8. **Widen transaction id node field from 12 to 14 bits** (review finding) -
   the transid layout put the API node id in bits 40-51, so node ids >= 4096
   carried into the block-number field and could produce duplicate
   transaction ids across different Ndb objects, corrupting TC's
   duplicate-transaction detection. Node id moved to bit 38 (good to 16383);
   verified safe against the 32-bit counter wrap and block-number-reuse
   carry-over logic, and nothing anywhere decodes those bits back out.
   *Revised by commit 13 (second review round, below): the bit-38 layout was
   not rolling-upgrade safe.*
9. **Bound automatic send buffer sizing by the number of CPUs** (review
   finding, the "32 GB issue") - the automatic memory configuration summed
   8 MB per configured transporter (64 GB at 8k slots). Now capped at
   32 MB per CPU with a 1 GB floor inside `Configuration::get_send_buffer()`
   (the single source for both the reservation and the
   RG_TRANSPORTER_BUFFERS limits); <= ~128-transporter clusters keep exactly
   the traditional sizing, and an explicit `TotalSendBufferMemory` is never
   capped. An 8k-slot config with default buffers and automatic memory now
   reaches STARTED on a 32 GB machine (previously OOM in start phase 0).
10. **Stop embedding node id ceiling sized arrays in every process** (review
    walkthrough) - three fixes: TlsKeyManager's ~1.6 MB cert table is now
    lazily allocated on first use (was embedded in every process incl. all
    tools); each trp_client's ~213 KB of per-Ndb-object send arrays are now
    grow-on-demand (typical clients settle at 16 entries); TRPMAN's ~730 KB
    per-receive-thread activity array is now runtime-sized from the
    configured max node id. The two items deferred here -
    `ClusterMgr::theNodes` (~1.5 MB/process) and
    `TransporterFacade::m_send_buffers` (~0.9 MB + 8192 embedded mutexes) -
    are completed in the second round, commit 16 below.

Second review round (review of commits 8-10 plus the two deferred
structures):

11. **Never throw from the trp_client send path when growing buffers** -
    fixes commit 10's grow-on-demand arrays: `getWritePtr()` is a
    TransporterRegistry callback with a defined out-of-resources contract
    (`SEND_BUFFER_FULL`), but growth used throwing `new`; on OOM a
    `std::bad_alloc` would unwind through NDB API code that releases
    `trp_client::m_mutex` by explicit unlock rather than RAII (leaked mutex,
    deadlock), and the second allocation could leak the first. Now
    non-throwing with the error reported through the existing retry path,
    plus explicit bounds asserts at the grown-array writes
    (`m_send_trps_list[cnt]` append, `updateWritePtr()`, `disable_send()`).
12. **Harden TlsKeyManager certificate table ownership** - the lazily
    allocated cert table made a second owning raw pointer freed in the
    destructor while copy operations were still implicit (double-free on an
    accidental copy): copies deleted; allocation on the TLS handshake path
    is now non-throwing (accessors already tolerate a null table); id-range
    guards enforced in release builds.
13. **Make the widened transaction id layout safe for rolling upgrade** -
    revises commit 8: moving the node id to bit 38 changed the prefix for
    *every* node id, so during a rolling upgrade a new-layout node N emitted
    exactly the prefix of old-layout node N/4 - two live API nodes could
    collide. The final layout keeps the released fields bit-identical below
    4096 and stores the two overflow bits in provably-zero spare bits:
    serial 0-31, spare 32-37, block-index bit 12 -> 38, node-id bit 12 -> 39,
    node id bits 0-11 -> 40-51, block-index bits 0-11 -> 52-63. This also
    closes a previously unchecked truncation: the full 16-bit block number
    was shifted into a 12-bit field, so two Ndb objects in one process with
    block indexes idx and idx+4096 (the facade permits 4711 clients)
    produced identical prefixes. `require()` now enforces both 13-bit fits.
14. **Guard every access to the runtime sized trp id arrays in ndbmtd** -
    hardens commits 4/10: `require(max_trp_id < glob_num_trp_ids)` where ids
    are registered (the sizing margin is exactly one transporter today, so a
    future extra transporter becomes a clean crash instead of silent heap
    corruption in the send path), bound guards in the recv-thread map
    functions (the epoll one moved above the map read), and a single checked
    accessor for TRPMAN's `m_trp_activity[]` covering all four sites. For
    the record: `m_trp_state[]`, `m_send_thread_instance_by_trp[]` and
    `g_trp_to_recv_thr_map[]` stay ceiling-sized (single instance, small).
15. **Read the CPU count for automatic send buffer sizing exactly once** -
    hardens commit 9: the reservation and RG_TRANSPORTER_BUFFERS
    computations now provably use the same CPU count even if cgroup CPU
    limits change between the calls; the cap message is logged once.
    Operational note (default change, from commit 9): a very large
    auto-sized config on a small-CPU host gets the cap (32 MB/CPU, 1 GB
    floor) instead of the summed per-transporter value; an explicit
    `TotalSendBufferMemory` is never capped.
16. **Allocate ClusterMgr nodes and facade send buffers per configured id** -
    completes the commit-10 walkthrough. Both structures become fixed arrays
    of atomic slot pointers (64 KB each); slots are heap-allocated once per
    configured id, fully initialized before a release-store publishes them,
    and never freed or moved until destruction - required because
    `theNodes` is read lock-free from hot paths (`getNodeInfo()`
    per-transaction node selection, `hb_received()` writing without a lock,
    references held across work) and `TFSendBuffer` embeds the mutex senders
    hold across `performSend()`. Access policy: registry/config-driven paths
    `require()` their slot; signal-driven handlers (NODE_FAILREP's unfiltered
    data-node bitmask, NF_COMPLETEREP, DUMP_STATE_ORD's signal-supplied ids,
    API_REGREQ/REGCONF/REGREF) bounds-check and skip ids outside the
    configuration view - the same policy that keeps data nodes from crashing
    on API_VERSION_REQ for unknown ids. The mgmd online reconfigure path is
    now transactional (allocate-all-then-publish, non-throwing) and
    growth-only: node additions apply online; node removals, node type
    changes and changes to the mgmd's own arbitration settings return
    failure into the existing 'this node need a restart' fallback instead of
    half-applying (behavioral change - the old in-place reset could clear a
    connected node's state without adjusting the connected/alive counters,
    and stopping the ArbitMgr from configure() would deadlock on the
    ClusterMgr lock). Net effect: ~2.3 MB saved per facade instance in every
    ndbapi process and mgmd, facade construction drops 8192 mutex
    initializations, and the 100 ms heartbeat scan touches 64 KB of pointers
    instead of a 1.5 MB sparse array. A flat pointer array was chosen over
    chunking for ownership simplicity; the ~128 KB residual is documented as
    a possible measured follow-up.

Mixed-version safety: while any connected or configured node lacks 8k support,
no id above 2039 can register and no high-id config can be fetched, so packed
bitmask sections never exceed 64 words in a mixed cluster. Existing
configurations are unaffected (`MAX_NODES` remains runtime-derived from the
actual config).

## Validation

* Full ndb MTR suite: 283 tests, all passing after triage (2 expected result
  updates; remaining failures were host-environment noise, detailed in the
  commits).
* Full multi-suite MTR coverage (all 18 RonDB-related suites in 26.02, per
  review guidance): ndb, ndb_binlog, ndbcluster, ndb_ddl, ndb_rpl, rpl_ndb,
  gcol_ndb, json_ndb pass on the dev box (2 tests pass serially only under
  parallel load); ndb_opt, ndb_quota, ndb_tls, ndb_ttl, ndb_ttl_rpl pass
  (100/100 in the combined run); ndbcrunch passes 18/18 after the suite fix
  in commit 6; ndb_ttl_purge structurally requires a WITH_RDRS build and now
  reports that per test (commit 5) instead of aborting the run; big-memory
  suites (ndb_big, rondb_big) were run on a 561 GB lab host - the
  feature-relevant tests (test_250_mysqld, ndb_many_tables) pass, remaining
  failures triaged to pre-existing harness defects present on 26.02-main
  (ConfigFactory does not expand ndb-cluster-connection-pool slots;
  DDL-vs-startup races; ndbinfo result files stale for the new
  rdma_transporters table) - reported separately, not patched here.
* mgmd allocation speedup A/B (commit 7; identical 3-mgmd cleartext rig,
  config and 4x16x64 client procedure as the multi-mgmd baseline below):
  **full 4096/4096 concurrent API connections in 27.6 min with 64/64 client
  processes surviving**, vs the 3776 plateau after ~47 min with 5 client
  deaths before; the 3776 mark is reached in ~16.5 min (2.9x); cross-mgmd
  reservation races dropped to zero and all three mgmds take allocation load
  concurrently from launch (previously one at a time); mass hard-kill of all
  4096 APIs cleaned up in <= 6 s (data node fds back to baseline), zero
  GCP-stop/watchdog events across the run.
* Unit: Bitmask TAP test extended with checked-decoder boundary cases
  (0/1/64/65/256/257-word sections).
* Distributed lab (physical hosts, 100/200 GbE): mgmd at node id 8190,
  2 ndbmtd, 7993 api slots (ids 100-8091, 8191), mysqlds at ids 8191/8090/8091
  and 100/101/102:
  - System start to STARTED in 15 s; config.bin 1.25 MB.
  - 80 concurrent DDL round-trips from both mysqlds simultaneously (high and
    low id tiers as coordinator+participant): 0 failures, 0 timeouts, full
    cross-visibility.
  - ~1050 concurrent API connections; DML sustained 6.2k-11.1k tx/s
    (99k inserts/s, 177k reads/s) through the full ramp and a simultaneous
    hard-kill of all connections; zero GCP-stop/watchdog events; ~4900
    disconnect cleanups processed.
  - Version gate verified against the live high-id cluster: a 26.02.8 client is
    cleanly refused at both `get config` and `get nodeid` (mgmd survives);
    26.02.9 is served.
* Multi-mgmd scale-out (same lab, 3 ndb_mgmd at ids 8188/8189/8190 + 2 ndbmtd,
  7993 api slots; 4 client hosts, 64 connections per client process, paced
  launch with per-host mgmd rotation):
  - **3776 concurrent API connections** held simultaneously (3.6x the
    single-mgmd result); paced allocation sustained 6.5-9.7 connects/s
    aggregate across the 3 mgmds vs ~1.7/s effective with one mgmd -
    `alloc_node_id` throughput scales with mgmd count.
  - Mass hard-kill of all 3776 APIs: cluster cleanup completed within one 28 s
    monitoring interval (data node fds 3836 -> 44), no errors, zero
    GCP-stop/watchdog/heartbeat events across the entire run.
  - mgmd RSS 1.9-2.6 GB per 0.7-2k held sessions at plateau, released back to
    ~0.6-0.7 GB after disconnect.
* Rolling upgrade/downgrade rehearsal with complete stock 26.02.8 binaries
  (mgmd + 2 ndbmtd + mysqld, continuous DML load throughout): upgrade order
  mgmd -> data nodes -> SQL nodes. At every mixed-version step the legacy
  paths were refused cleanly: old-client config fetch, high-id and low-id
  `get nodeid` (instant protocol errors), and an old-ndbd restart against a
  high-id config (refused at both config fetch and allocation; cluster
  unaffected; the node rejoined after the config was reverted). Mixed
  data-node versions served SQL + DML with zero GCP events. After high ids
  were enabled, the upgraded mysqld registered at id 8191 and ran DDL, and
  clients connected at ids 2040/8190. Full downgrade back to stock 26.02.8
  succeeded with all data intact, including a table created at id 8191.
* TLS at scale (same lab, `RequireTls`+`RequireCertificate` on data nodes =
  TLS 1.3 on every transporter link; keys via ndb_sign_keys): **full 4096/4096
  concurrent TLS API connections** (8192 encrypted links across 2 data nodes),
  all holder processes survived; certless client cleanly refused; zero
  GCP/watchdog events; mass hard-kill of all 4096 cleaned up in <= 14 s
  (data node fds 4139 -> 43). NDB API clients pick up link TLS automatically
  from the compiled-in default search path with no code changes.
* Sanitizers: Bitmask checked-decoder unit test passes under ASan+UBSan; a
  pure-ASan cluster (mgmd at id 8190 + ndbmtd) ran connects at
  2039/2040/2047/2048/4095/4096/8191 (all register and execute) and refused
  8192 with a clean "illegal nodeid" - zero server-side sanitizer reports.
  Only finding: a pre-existing stack-use-after-return in the flexAsynch test
  tool's receive path, firing at all node ids including pre-change ids -
  unrelated to this work, flagged separately.
* Small-cluster memory A/B (identical 1 mgmd + 1 ndbmtd + 5 api slot config,
  stock 26.02.8 vs this branch): before the sizing fix the 8k ceiling cost
  +338 MB idle ndbmtd RSS (1.44 -> 1.81 GB; global memory pools byte-identical,
  all growth in ceiling-scaled runtime arrays). With the fix: **1.39 GB - below
  the stock baseline**, since the runtime bound (297 slots at max node id 200)
  also beats stock's 2136-slot compile-time ceiling. A config using node id
  8191 sizes to the full 8288 slots and boundary connects at 2040/8191 pass.
* API-client/mgmd memory A/B (commit 10, same 5-slot rig): single-connection
  API client RSS 18.4 -> **10.7 MB - below the 13.2 MB stock baseline**
  (grow-on-demand send arrays also beat stock's 2040-slot embedded arrays);
  ndb_mgmd 17.4 -> 11.4 MB (remaining +1.1 MB vs stock attributed to the
  `theNodes`/facade items, since addressed by commit 16 - re-measurement
  pending, see below). ndbmtd additionally saves ~730 KB per receive thread
  at scale. Validated with ndb_trpman-t (22307 assertions), ndb_basic,
  ndb_high_nodeid and ndb_tls.tls_off_certs.
* Second review round (commits 11-16): clean build with no new warnings in
  the touched files; TlsKeyManager-t and ndb_trpman-t (22307 assertions)
  pass; MTR ndb.ndb_basic, ndb.ndb_high_nodeid (transactions at node id
  8191, exercising the revised transid layout), ndb.ndb_addnode and
  ndb.ndb_addnode_restart (the mgmd online-reconfigure growth path end to
  end, exercising the new stable-slot publication) and ndb_tls.tls_off_certs
  all pass.
* Pending rig availability (will be appended here when run): 5-slot rig
  memory A/B rerun for commits 11-16 (expected: API client ~10.7 -> ~9.2 MB,
  ndb_mgmd ~11.4 -> ~9.4 MB, dropping mgmd below the 10.3 MB stock
  baseline); 7997-slot startup sanity rerun on the 32 GB box; an ASan MTR
  pass covering the new heap ownership in long-lived objects; an online
  config change that *removes* a connected node (now expected to log and set
  'this node need a restart' rather than half-apply).
* Send-buffer CPU cap (commit 9) verified with the exact previous failure
  case: 1 mgmd + 1 ndbmtd + 7997 api slots, DEFAULT per-transporter buffers,
  AutomaticMemoryConfig - the derived 63992 MB send buffer is capped to
  1024 MB (16 CPUs, logged with the override hint) and the node reaches
  STARTED on a 32 GB machine; previously OOM-killed in start phase 0.
* Transid field fix (commit 8) exercised by ndb_high_nodeid (DDL + DML
  through mysqlds at node ids 8191/2040/8000, i.e. transaction ids built
  with the 14-bit node field above the old 4095 boundary).
* Boundary ids exercised: 2039/2040/2047/2048/4096/8000/8190/8191.

## Operational notes for high slot counts (documented findings)

* Default per-transporter buffers (2M+2M) at thousands of configured slots:
  the send-buffer side is now handled by the CPU cap in commit 9 (an 8k-slot
  default-buffer config starts on a 32 GB machine); receive buffers are
  still eagerly allocated per configured slot (~64 KB floor per slot at
  explicit small settings), so bounded `ReceiveBufferMemory` settings remain
  recommended for very high slot counts.
* A fixed `ServerPort` is required (otherwise each node pair binds its own
  dynamic listener -> fd exhaustion), and fd limits must cover the configured
  slot count.
* mgmd memory during fleet connect scales with sessions (~1.8-3.5 MB/session
  transient with MB-size packed configs); fleet cold-start rate was bounded
  by the serialized `alloc_node_id`+config-fetch path (~5-8 connects/s
  measured). Commit 7 removes the allocation-side serialization (8-11.5
  connects/s sustained through the ramp, full 4096 reached); the remaining
  per-session config-fetch cost (node-filtered packing) stays listed as
  follow-up. The next structural step, if ever needed, is a concurrent
  allocation-op pool in the QMGR master (today a single in-flight allocation
  slot cluster-wide) - protocol-compatible since mgmds already handle Busy.
* ndb_mgmd needs the same fd-limit preflight as data nodes: every connected
  API holds its MGM session open, so mgmd sessions/fds scale with the
  connected fleet (a test run with the default 1024-fd soft limit wedged at
  exactly ~1022 sessions per mgmd while it burned CPU on accept/refuse
  churn).
* Client retry storms cause congestion collapse on a backlogged mgmd: clients
  retrying every 5 s generate ~120 session-churns/s (mgmd at ~2 CPU cores,
  mostly config-fetch work) and effective allocation drops below 1/s until the
  backlog drains (2.7x alloc-request amplification measured). Large fleets
  should raise client connect budgets/backoff (`ConnectBackoffMaxTime`) and
  pace cold-start ramps.
* First-time start of multiple mgmds with `--initial` must be sequential; a
  simultaneous start can collide in the config-change protocol (one mgmd exits,
  survivors keep serving the last good config but refuse further config
  changes).
* Enabling ids > 2039 fences out still-connected old API/SQL nodes at their
  next re-registration (clean 4038/4009 errors, graceful retries, no crashes) -
  upgrade all API/SQL nodes before extending the config. Downgrade requires two
  separate rolling restarts: first revert the config to ids <= 2039 on the new
  binaries, then downgrade binaries; a combined restart is refused by peers
  still running the high-id config (QMGR error 6304, clean node exit).

### DIFF
--- a/storage/ndb/src/ndbapi/ClusterMgr.cpp
+++ b/storage/ndb/src/ndbapi/ClusterMgr.cpp
@@ -26,6 +26,7 @@
 
 #include <ndb_global.h>
 #include <ndb_limits.h>
+#include <new>
 #include <util/version.h>
 
 #include <kernel/GlobalSignalNumbers.h>
@@ -116,6 +117,7 @@ ClusterMgr::ClusterMgr(TransporterFacade & _facade):
   noOfConnectedDBNodes(0),
   minDbVersion(0),
   minApiVersion(0),
+  m_configured(false),
   theClusterMgrThread(nullptr),
   m_process_info(nullptr),
   m_cluster_state(CS_waiting_for_clean_cache),
@@ -126,6 +128,9 @@ ClusterMgr::ClusterMgr(TransporterFacade & _facade):
   m_node_change_count(0)
 {
   DBUG_ENTER("ClusterMgr::ClusterMgr");
+  for (Uint32 i = 0; i < ABS_MAX_NODES; i++) {
+    theNodes[i].store(nullptr, std::memory_order_relaxed);
+  }
   clusterMgrThreadMutex = NdbMutex_Create();
   m_node_state_mutex = NdbMutex_Create();
   waitForHBCond= NdbCondition_Create();
@@ -149,6 +154,15 @@ ClusterMgr::~ClusterMgr() {
     delete theArbitMgr;
     theArbitMgr = nullptr;
   }
+  /**
+   * The receive/send/ClusterMgr threads were stopped before we get
+   * here (TransporterFacade::stop_instance()), so the node slots can
+   * now be freed: this is the only place they are ever deleted.
+   */
+  for (Uint32 i = 0; i < ABS_MAX_NODES; i++) {
+    delete theNodes[i].load(std::memory_order_relaxed);
+    theNodes[i].store(nullptr, std::memory_order_relaxed);
+  }
   NdbCondition_Destroy(waitForHBCond);
   NdbMutex_Destroy(clusterMgrThreadMutex);
   NdbMutex_Destroy(m_node_state_mutex);
@@ -157,12 +171,56 @@ ClusterMgr::~ClusterMgr() {
 }
 
 /**
- * This method is called from start of cluster connection instance and
- * before we have started any socket services and thus it needs no
- * mutex protection since the ClusterMgr object isn't known by any other
- * thread at this point in time.
+ * ::configure()
+ *
+ * Called in two different contexts:
+ *
+ * 1) Initial start of the cluster connection instance
+ *    (TransporterFacade::start_instance()), before any socket services
+ *    have been started: the ClusterMgr object isn't known by any other
+ *    thread yet, so no mutex protection is needed.
+ *
+ * 2) mgmd online configuration change (MgmtSrvr::config_changed() ->
+ *    TransporterFacade::configure()) while the ClusterMgr thread, the
+ *    receive thread and MGM sessions are live and reading theNodes[]
+ *    (mostly lock-free). For this path the change is:
+ *    - growth-only: node additions are applied online; node removals,
+ *      node type changes and changes to our own arbitration settings
+ *      are refused by returning false, which MgmtSrvr answers with its
+ *      existing 'this node need a restart' fallback (m_need_restart);
+ *    - transactional: all new node slots are allocated (non-throwing)
+ *      and fully initialized before anything is published; on
+ *      allocation failure nothing has become visible and false is
+ *      returned;
+ *    - safely published: a slot pointer is release-stored only after
+ *      the Node is completely initialized (including 'defined'), so
+ *      the lock-free acquire-loaders never observe a half-built node.
+ *      Slots are never freed or moved again (readers keep plain
+ *      references across their work, see theNodes[]).
+ *    In-place updates of already published nodes (the m_node_active
+ *    flag) are done under the ClusterMgr lock like every other writer.
+ *    Note that this serializes the *writers*; the long-standing
+ *    lock-free *reads* of node state remain exactly as before.
  */
-void ClusterMgr::configure(Uint32 nodeId, const ndb_mgm_configuration *config) {
+bool ClusterMgr::configure(Uint32 nodeId, const ndb_mgm_configuration *config) {
+  const bool live_reconfig = m_configured;
+
+  /**
+   * Phase 1: parse the node sections into a plan, without side effects.
+   * plan[nodeId].defined == 0 means the node id is not in the new
+   * configuration.
+   */
+  struct NodePlan {
+    Uint8 defined;
+    Uint8 has_type;
+    Uint8 type;  // NodeInfo::NodeType value, valid when has_type == 1
+    Uint8 active;
+  };
+  NodePlan *plan = new (std::nothrow) NodePlan[ABS_MAX_NODES]();
+  if (unlikely(plan == nullptr)) {
+    return false;
+  }
+
   ndb_mgm_configuration_iterator iter(config, CFG_SECTION_NODE);
   for (iter.first(); iter.valid(); iter.next()) {
     Uint32 nodeId = 0;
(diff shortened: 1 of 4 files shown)


elemberger-peak:
@mikaelronstrom — following up on your review feedback (small-cluster memory efficiency and the full ~18-suite MTR coverage). Both are now addressed in this PR, and the testing surfaced a few pre-existing issues you should know about — most are upstream/26.02-main items, one I fixed here since it was blocking the coverage runs.

### 1. MTR coverage status

All 18 RonDB-related suites are now accounted for: everything passes except items triaged to documented pre-existing causes (details in the PR description). The big-memory suites (ndb_big/rondb_big) need a large-RAM host — a 32 GB dev box gets OOM-killed by design of those tests. ndb_ttl_purge structurally requires a WITH_RDRS build (its my.cnf defines rdrs processes); with the harness fix in this PR it now reports that per-test instead of aborting the whole run.

### 2. Fixed in this PR: ndbcrunch has been broken since RONDB-210 (2023)

Commit 589a4c4b759 commented out `[mysqld.1.crunch]`/`ndb-log-bin=1` in the suite's my.cnf. That had two effects: `ndb_log_bin=0` everywhere (nothing ever binlogged), and — more subtly — `[mysqld.2.crunch]` became the first mysqld group, so it took the first port and became mysqltest's default connection. `replicate_mysqld2.inc` then saved the "source" binlog position from the applier and waited forever for positions from two different servers to match, so every crunch_update* test ran into the 900 s timeout. The post-timeout SIGABRT core dumps make it look like data node crashes, but the cluster is healthy. Fixed in this PR (group restored with the ordering requirement documented, plus test-side replication cleanup); all 18 ndbcrunch tests pass and the recorded .result files are unchanged.

### 3. mgmd node id allocation under mass API connects (also in this PR)

While storm-testing thousands of concurrent API connections I traced the fleet cold-start bottleneck to the allocation path: QMGR's master has a single in-flight allocation slot (`opAllocNodeIdReq`) and the mgmd retries Busy on a fixed 100 ms sleep; on top of that, every mgmd scans wildcard candidates from the lowest free id, so multiple mgmds constantly race for the same id (NodeReserved round trips). Two mgmd-local, protocol-compatible changes: adaptive backoff (2 ms first retry, jittered, capped at the old ~100 ms under sustained contention) and a rotated scan start (own-node-id hash + round robin, only for API allocations with >= 64 equal candidates, so small clusters keep deterministic assignment and existing tests are byte-identical).

A/B on the same 3-mgmd / 7993-slot / 4096-client rig: full **4096/4096 connected in 27.6 min with zero client deaths**, vs a 3776 plateau at ~47 min with 5 dead clients before; cross-mgmd reservation races went to zero.

If we want to go further, the next structural step would be a small pool of concurrent allocation operation records in the QMGR master (keyed by node id) instead of the single slot — protocol-compatible since mgmds already handle Busy. Happy to prototype if you think it's worth it; after the mgmd fix the practical limiter is per-session config fetch (the node-filtered packing follow-up) and client-side retry pacing.

### 4. Pre-existing issues found while testing (not touched in this PR)

- **mysql-test-run.pl hangs forever after a testcase timeout**: it SIGABRTs the ndbmtd to get a core, then waits indefinitely for a process that is wedged in its crash handler; the whole run stalls until killed by hand.
- **ConfigFactory.pm does not auto-expand API slots for `ndb-cluster-connection-pool`** like upstream's: ndb_big.bug13637411 sets pool=49 with only 2 API slots in the generated config, so mysqld can never become ready ("No free node id found for mysqld(API)" forever) and the test times out. crund/bug37983/rqg_spj/ndb_big_addnode show related startup-availability signatures, plus a DDL-vs-startup race (CREATE TABLE issued ~3 s after start-phase completion hits NDB 4039 with DataMemory=21G).
- **ndbinfo + ndbinfo_plans .result files are stale on 26.02-main** since the rdma_transporters ndbinfo table was added (table count 56 -> 57); fails on a clean checkout regardless of this PR.

### 5. Ops note worth documenting

ndb_mgmd needs the same fd-limit guidance as data nodes at high slot counts: every connected API holds its MGM session open, so mgmd sessions/fds scale with the connected fleet. With the default 1024 soft limit a storm wedges at ~1022 sessions per mgmd while it burns CPU on accept churn.

Everything above (numbers, procedures, per-suite tally) is in the PR description. Let me know if you want the QMGR allocation-pool prototype or if you'd rather keep that for a separate discussion.

elemberger-peak:
@mikaelronstrom — all three review items are now addressed and pushed:

**1. Transaction id 12-bit node field** (commit 08b08c14609): shifted the node id to bit 38 as you prescribed — 14-bit field, good to 16383. Before changing it I verified the full lifecycle: the serial-counter wrap in `startTransactionLocal()`/`allocate_transaction_id()` only cycles the low 32 bits and `get_next_transid()`/`set_next_transid()` carry a `Uint32`, so bits 32-37 are provably always zero and the widened field collides with nothing. I also extended your kernel-blocks check to the whole `src` tree — nothing anywhere decodes bits ≥ 32 back out of a transid. A comment documents the layout at the construction site. `ndb_high_nodeid` (transactions at node id 8191) passes.

**2. The 32 GB issue** (commit 7791c97e8e9): the automatic sizing lives in `Configuration::get_send_buffer()`, which turned out to be the single choke point used by both the automatic-memory reservation and the `RG_TRANSPORTER_BUFFERS` limits, so one cap keeps both consistent. The derived value is now capped at 32 MB per CPU with a 1 GB floor (≤ ~128-transporter clusters keep exactly the traditional sizing; explicit `TotalSendBufferMemory` is never touched). Verified with the real failure case: 7998 transporters × defaults = 63,992 MB → "capped to 1024 MBytes based on 16 CPUs", and the 8k-slot config with *default* buffers now reaches STARTED on a 32 GB machine — previously an instant OOM in start phase 0. Happy to adjust the 32 MB/CPU and 1 GB constants if you prefer different values.

**3. ABS_MAX_NODES walkthrough** (commit 74047849dd0): full inventory done; three structures made normal deployments pay materially and are fixed: `TlsKeyManager::m_cert_table` (~1.6 MB in every process, now lazy), per-Ndb-object `trp_client` send arrays (~213 KB each, now grow-on-demand — a mysqld with hundreds of Ndb objects previously paid tens of MB), and `Trpman::m_trp_activity` (~730 KB per receive thread, now runtime-sized like the earlier mt.cpp fix). Measured on the minimal 5-slot cluster: API client RSS 18.4 → 10.7 MB and mgmd 17.4 → 11.4 MB — the client is now *below* the pre-8k 26.02.8 baseline (13.2 MB), since grow-on-demand also beats the old 2040-slot arrays. Two items are documented rather than fixed: `ClusterMgr::theNodes` (~1.5 MB/process) and `TransporterFacade::m_send_buffers` (~0.9 MB plus 8192 embedded mutexes/process) — both need a design that survives configuration growth on reconfigure (they are indexed by node/trp ids that can grow), so I'd rather do those as a considered follow-up than rush them. The kernel block arrays (Dbtc/Dbtup/Suma/Qmgr etc.) are 16-64 KB each and left alone.

Details and numbers are in the PR description (commits 8-10).

elemberger-peak:
Pushed the second review round as commits 11-16 (`19f5e7952cd..a304fdb20fb`); the PR description has a new "Second review round" section with the details. Highlights:

- The two items deferred from the memory walkthrough are now done: `ClusterMgr::theNodes` and `TransporterFacade::m_send_buffers` are stable-slot tables (atomic pointers, allocated per configured id, published complete, never moved). The mgmd online reconfigure path is transactional and growth-only; removals/type/arbitration changes now fall back to the existing "this node need a restart" path instead of half-applying.
- The transid widening is revised to a layout that is bit-identical to the released one below node id / block index 4096, removing the rolling-upgrade collision, and now also covers the block-index overflow with `require()`s.
- Send-path OOM hardening (no throwing `new` in `getWritePtr()`), TlsKeyManager ownership fixes, runtime-bound guards for all trp-id indexed arrays in ndbmtd, and a single cached CPU count for the send-buffer cap.

Validated so far: clean build, TlsKeyManager-t + ndb_trpman-t, MTR ndb_basic / ndb_high_nodeid / ndb_addnode / ndb_addnode_restart / ndb_tls.tls_off_certs all green. The 5-slot memory A/B rerun, 7997-slot startup sanity and an ASan MTR pass are queued for the next rig slot and will be appended to the description.

mronstro:
Most of the last commits looks like good commits, haven't had time to analyse them in more detail. But looks like moves in the right direction. Last commit needs some more careful analysis.

PR #1044, opened 2026-08-14
Bugfix backup system

KernelMaker:


KernelMaker:
## Fix 1/6 — LCP file write errors reported as "internal program error" (2341)

**Commit:** 97781f76d08 *Fix LCP file write errors being reported as internal program errors*

### Issue
A file system error (e.g. disk full, 2810) on an LCP data file write killed the node with error 2341 "Internal program error" instead of the real cause, because the BACKUP block closed the failed file while the fragment scan still owned it:

- `execFSAPPENDREF` cleared only `BF_FILE_THREAD` and `checkFile()` closed the file immediately, so `execFSCLOSECONF` failed its flags `ndbrequire` with `BF_SCAN_THREAD` still set.
- In a multi-file fragment LCP the error can land on any data file while `dataFilePtr[0]` stays clean, in which case `finalize_lcp_processing()` failed `ndbrequire(filePtr.p->errorCode == ptr.p->errorCode)` instead.
- A late SCAN_FRAGREF could report the scan error and mask the file system error recorded on another data file.

### Solution
- `checkFile()` defers the close of a failed LCP data file while `BF_SCAN_THREAD` is set; the scan detects the error on its next interaction and terminates the fragment.
- `fragmentCompleted()` detects errors across *all* data files of the fragment LCP, detaches the scan from all of them, and sends BACKUP_FRAGMENT_REF with the failing file's error.
- `find_lcp_error_file()` makes every report site (`execSCAN_FRAGREF`, `fragmentCompleted`, `finalize_lcp_processing`) agree on the first recorded error; `finalize_lcp_processing` no longer assumes `dataFilePtr[0]` carries it.

The failure remains fatal by design (DBLQH calls `progError` on BACKUP_FRAGMENT_REF), but the node now dies reporting the real error: "Unable to store fragment during LCP. NDBFS Error: `<code>`".

### Risk
Low. All changes sit on the LCP file-error path; healthy-path behavior is untouched, and the failure stays fatal, so no new survival semantics are introduced. The `TableNotDefined`/`ActiveSchemaTrans` path records no file error and is unchanged.

**Test:** new error insert 10057 (non-crashing FSAPPENDREF while the scan is active — the existing 10044/10045 inserts crash before the close path runs, which is why this was never covered); new testNodeRestart case `TestLCPFSErrScanActive` asserts the victim dies with the injected error 2812 instead of 2341 (red before / green after). Registered in daily-devel--02.

KernelMaker:
## Fix 2/6 — late LCP write errors masked by the row-count check

**Commit:** bdee0b060be *Fix late LCP file write errors being masked by the LCP row count check*

### Issue
A file system error on an LCP data file write arriving *after* the fragment scan completed was still reported as 2341 whenever the fragment's row counts were also inconsistent: only the scan completion path assigns `m_save_error_code`, so the row-count `ndbrequire` in `lcp_write_ctl_file` ran with no record of the file error and fired before `finalize_lcp_processing` could report the real error to DBLQH. In addition, `m_save_error_code` was never initialized per fragment, so error paths could read the previous fragment's value — or garbage on first use.

### Solution
- `checkFile()` records the file error in `m_save_error_code` (first-wins) when the LCP slave state is STOPPING — the only window that reaches the row-count check, entered right after scan completion assigns the field.
- `start_execute_lcp()` initializes `m_save_error_code` per fragment.

### Risk
Low. The change only records an error code earlier on an already-failing path; there is no control-flow change for healthy LCPs. The per-fragment initialization removes a latent stale/garbage read.

**Test:** new error inserts 10058 (non-crashing FSAPPENDREF after scan completion) and 10059 (same, plus a row-count skew on the errored fragment that provably falsifies both disjuncts of the check); new testNodeRestart cases `TestLCPFSErrAfterScan` and `TestLCPFSErrAfterScanSkew` — red 2341 at the row-count `ndbrequire` before the fix, green 2812 after. Registered in daily-devel--02.

KernelMaker:
## Fix 3/6 — fragment scan errors crash the backup master and the erroring worker

**Commit:** 06d4dcb5256 *Fix backup fragment scan errors crashing the master and the worker* (crash half; fix 4/6 covers the error-attribution half of the same commit)

### Issue
BACKUP_FRAGMENT_REF carries no table/fragment identity, and the master's match loop consumed the *first* in-flight fragment of the REF's node. With multithreaded backup (several LDM workers of one node scanning concurrently) a worker's REF could consume **another worker's** fragment: that fragment was marked scanned while its scan was still running, and the master died on the `scanned == 0` check when the real CONF arrived — or, when several REFs each consumed a wrong slot, the emptied send counter stopped the backup under an active scan.

The erroring worker had a second crash in the same scenario: `checkFile()`'s error branch closed the file during the stop phase without clearing `BF_FILE_THREAD`, tripping FSCLOSECONF's file-flags check; for a backup not yet stopping, the file thread ended silently with the flag set and `closeFiles()` would wait for it forever.

### Solution
- The master now identifies the fragment by **(node, LDM instance)**: a worker sends its REF directly from the LDM instance owning the fragment, so the match additionally requires `mapFragToLdm()` to equal the sender's instance. Node-failure handling instead fabricates REFs to self on behalf of dead nodes (sender node != `ref->nodeId`); those keep the old consume-any-fragment behavior, since a dead node's fragments must all be released.
- The worker's `checkFile()` error branch now ends the file thread explicitly, so FSCLOSECONF sees consistent flags and `closeFiles()` is never left waiting.

### Risk
Low–medium: the master's matching logic changes on the REF path, but a REF that fails the stricter match now falls through to the unmatched-REF handling (fix 4/6) instead of corrupting the fragment bookkeeping, so a mismatch can no longer stop or crash the backup. All changes are receive-side, so mixed-version clusters are safe.

**Test:** error inserts 10060/10061 fail the backup scan of one selected fragment mid-scan; testBackup `FragmentRefWrongWorker` asserts the master survives, the abort reports the worker's real error, and a subsequent backup succeeds.

KernelMaker:
## Fix 4/6 — unmatched fragment REF fabricates LogBufferFull and hides the real error

**Commit:** 06d4dcb5256 *Fix backup fragment scan errors crashing the master and the worker* (error-attribution half; fix 3/6 covers the crash half of the same commit)

### Issue
A BACKUP_FRAGMENT_REF matching no in-flight fragment discarded the worker's real error code and fabricated **LogBufferFull (1324)** as the abort reason. A backup failing on, say, a file system error could therefore be diagnosed as a `BackupLogBufferSize` tuning problem, sending operators down the wrong path — and before fix 3/6, duplicated or mis-matched REFs made this window reachable in practice.

### Solution
A REF matching no in-flight fragment now:
- records the worker's **real** error code,
- leaves the send counter alone (it holds no reservation for an unmatched REF), and
- (re-)triggers the abort with `FileOrScanError` instead of fabricating LogBufferFull.

### Risk
Low. Only the unmatched-REF path changes; the backup still aborts, now with truthful error attribution. All receive-side, mixed-version safe. Practical consequence: after this fix, a 1324 abort code can be trusted as a genuine log-buffer condition when sizing `BackupLogBufferSize`.

**Test:** testBackup `FragmentRefDuplicate` (error insert 10061 duplicates the REF) asserts the master survives the duplicate, the abort carries the worker's real error, and a subsequent backup succeeds.

KernelMaker:
## Fix 5/6 — failed backups leave their files behind on the errored node

**Commit:** 42e2fd8b458 *Fix failed backups leaving their files behind on the errored node*

### Issue
A backup that failed on a file error left its `BACKUP-<id>` files on disk forever, silently filling the backup volume. Two defects stacked:

- `cleanupNextTable()` zeroed `backupId` to `~0` **before** `removeBackup()` stamped it into the FSREMOVEREQ, so the removal targeted the nonexistent `BACKUP-4294967295` and "succeeded" without removing anything.
- For mid-flight aborts `removeBackup()` never ran at all: the stop path clears `errorCode` so that STOP_BACKUP confirms an already-reported failure, and cleanup's `checkError()` then saw no error.

### Solution
- The `backupId` reset moves to the release points, after the removal has used the real id.
- A new flag, `m_remove_files_after_cleanup`, carries the failure across the `errorCode` reset so cleanup knows to remove the files.
- Removal is scoped to what the attempt **provably owns**: files are opened with `OM_CREATE_IF_NONE`, so an open CONF means this attempt created the file, and a per-REF `fsErrFileExists` means a fileset with this id already existed — in that case cleanup removes nothing, protecting a valid earlier backup under a reused id. The removal unlinks exactly the attempt's three files by replaying their open-time addressing (leaf filenames carry the node id), so it is safe across shared `BackupDataDir` setups, node id changes, and single-threaded/mt layout changes.

### Risk
The inherent risk of any cleanup fix is removing valid files; it is mitigated by the ownership guards above and by using no directory-level removal at all (the emptied directory shells are left behind here and handled by fix 6/6). Also hardened in this commit: FSREMOVEREQ's `fileNumber` words are fully initialized (a stale part number would redirect the path into a part subdirectory), `AbortBackupOrd::senderRef` is no longer sent uninitialized, and error insert 10036 is now backup-only (without an `is_lcp()` guard, an LCP write racing the armed insert took error 2810, which DBLQH treats as fatal for the node).

**Test:** new testBackup case `FailedBackupRemovesFiles` (daily-basic--04): clean verified baseline, EI 10036 on a non-master node, the backup must fail with abort error 2810, the errored node's backup file content must disappear, the cluster must stay up, and a final successful backup proves the insert was cleared.

KernelMaker:
## Fix 6/6 — abort-protocol cleanup: failed backups clean up on every participant, dead participants are swept on rejoin

**Commits:** a3498246bb6, e1ba0105960, 6edf7db37c9, 5db7e55f04a — four commits forming one protocol change. Fix 5/6 removes a failed backup's files on the node that saw the error; these commits close the remaining ways debris (or worse) survives a failed backup.

### Issue
- **a3498246bb6 — a leaked backup record kills the node on every following backup.** A backup record that is never released (e.g. a master-death abort stalling before cleanup completes) permanently occupies the node's backup pool slot. The next backup's DEFINE_BACKUP_REQ failed the seize and hit `ndbabort` ("If master has succeeded slave should succeed", error 2301) — one stalled abort cascaded into node death on every subsequent backup attempt.
- **e1ba0105960 — aborted backups leave files on non-errored participants.** An aborted backup only removed files on workers that saw an error themselves: STOP_BACKUP_REQ carries no abort reason and `abort_scan` only reaches workers with an active scan. With an error arriving in the stop phase, every worker except the erroring one leaked its partial files — including the master node's other LDM workers.
- **6edf7db37c9 — cleanup leaves empty BACKUP directory shells behind.** Scoped removal deletes exactly the three files, because NDBFS's only directory removal was a recursive `rm -rf` — unsafe on a shared `BackupDataDir`. The emptied `BACKUP-<id>` and mt `BACKUP-<id>-PART-N-OF-M` directory shells accumulated forever.
- **5db7e55f04a — dead participants keep a failed backup's files after restart.** A participant that dies during a backup never runs its own cleanup, and nothing removed its partial files when it came back: the debris survived every restart.

### Solution
- **Leaked record:** fail the backup instead of the node — log the stuck record's state and reply DEFINE_BACKUP_REF with new error 1352 (`FailedToAllocateBackupRecord`). Abort orders whose `backupId` does not match the record are ignored, so the failed backup's abort cannot alias and resurrect the stale record. DBLQH's one-time LCP define keeps its `ndbabort` (an occupied LCP slot is unrecoverable state corruption).
- **Cleanup on all participants:** on the error path the master fans the new `AbortBackupOrd::CleanupFailedBackup` in place of BackupComplete; every receiver runs the failed-backup file removal with the same ownership guards as fix 5/6.
- **Directory shells:** FSREMOVEREQ gains a non-recursive, empty-directory-only mode (a new ninth signal word, read only when carried, so all existing senders keep their behavior; node-local signal, no version gating). Each worker removes its own mt part directory, then attempts the shared `BACKUP-<id>` parent only after its own part removal confirmed; a directory still holding anything is left untouched, and a removal failure degrades to leaving that shell — it never blocks the record release.
- **Dead participants:** the cluster master's LDM1 backup worker records what a dead participant owes (latest failed backup per node, with layout and a generation tag) — only once the backup provably passed DEFINE on that node, and never after the backup's success verdict, so a valid older fileset under a reused id is never touched. On rejoin, a version-gated `AbortBackupOrd::RemoveFailedBackupFiles` makes the node sweep exactly the failed attempt's layout. The entry is retired only by a generation-matched confirm (a node dying mid-sweep is re-ordered on its next rejoin). On the receiving node the order passes through the BackupProxy, which bars any same-id DEFINE while the sweep runs (new error `FailedForDebrisSweepInProgress`) — a define racing the removals could otherwise leave two removers over the same paths.

### Risk
- **Version gating / mixed clusters:** the new inter-node orders are gated per receiving node on the capability `ndbd_backup_failed_cleanup_ord` (floor **25.10.18**; forward-port floors 26.2.10/26.4.2/26.5.1 are placeholders for the ports). Nodes below the floor receive exactly the old signals and keep the old behavior, so rolling upgrades are safe. **Release coupling: on this 25.10.17 tree the mechanism is dormant — the release that ships this PR must bump MYSQL_VERSION to 25.10.18 for it to activate.** The four gated tests (`AbortRemovesAllNodesFiles`, `DeadParticipantSwept`, `DeadParticipantSweptAtStop`, `SweepBlocksSameIdBackup`) self-skip until then.
- **File safety:** all removals are the same scoped, ownership-guarded unlinks as fix 5/6, plus rmdir-only directory removal that tolerates non-empty/missing directories — nothing recursive, nothing outside the failed attempt's recorded layout.
- **Accepted residuals (by design):** the proxy's sweep barrier has no timeout — an unresponsive NDBFS keeps it up, since clearing it on a timer could let a define race the sweep; a node that declines a sweep order retries only on its next restart (conservative boot-lifetime exclusion); the sub-millisecond window between a backup's success verdict and its recording is guarded by the verdict flag plus generation-floored cancellation, but is not black-box testable.

**Test:** `StaleRecordFailsDefine` (EI 10062 leaks the record at a successful backup's cleanup; following backups must fail with 1352, no node death), `AbortRemovesAllNodesFiles` (EI 10063 fails the stop phase; no node may hold `BACKUP-<id>` files afterwards), `DeadParticipantSwept` (mid-scan and stop-phase variants, plus a fabricated valid fileset that must survive a sweep under a reused id), and `SweepBlocksSameIdBackup` (EI 10064 holds the sweep window open; a same-id backup must fail with `FailedForDebrisSweepInProgress`, with every non-LDM1 worker armed to crash if the define ever reaches it).

PR #1055, opened 2026-08-26
RONDB-1111: Fix issues with TC takeover crash and hangs

mronstro:


KernelMaker:
I think the 26.x version gate needs another look before this merges.

`ndbd_take_overtcconf_rebroadcast()` considers every node at 26.2.10 or newer safe to receive these rebroadcasts. However, current `26.02-main` is already 26.2.10 and current `26.04-main` is 26.4.2, and both still have the old `ndbrequire(tcFailNo < senderTcFailNo)` path in `execTAKE_OVERTCCONF`.

That makes the mixed-version case unsafe. If one takeover has completed and the master then fails, a 25.10.18 node becoming the new master will re-announce the completed takeover to those 26.x peers. A peer that received the original CONF has already removed its queue entry, so the duplicate reaches the old `ndbrequire` and crashes the node—the failure this change is intended to prevent.

Could we either forward-port the receiver-side duplicate handling to every 26.x version advertised by this gate before shipping, or make the gate per-series and set each floor to the first release that actually contains the fix? The single `>= 26.2.10` comparison also opts every later 26.x line in automatically, so those branches need to carry the receiver change too.

PR #1073, opened 2026-09-03
RONDB-1058 follow-up: Re-enable the arbitrator startup gate in Arbitr…

mronstro:
…atorGateNotReady

Commit 97836113b5a disabled the arbitrator startup gate for all testMgmd subtests by letting ConfigFactory put ArbitrationRankWait=0 in every generated data node section. That also disabled it for the one subtest that exercises the gate: ArbitratorGateNotReady failed with CHECK(observed_gate) after three "Missed the gate window" restarts, since wait_until_arbitrator() returned on its rankWait==0 early exit and every status poll succeeded.

Restore the default ArbitrationRankWait for the (never started) data node in runTestArbitratorGateNotReady, so the 3 second cold-start gate window the test relies on is armed again.

ConfigFactory::put() silently failed when the key already existed in the section, because Properties::put() defaults to replace=false. Pass replace=true so a test can override a value create() already set. Existing callers only add new keys and are unaffected.

Claude-Session: https://claude.ai/code/session_01DCZz6MFMiRgXR5uV9G2YPQ

### DIFF
--- a/storage/ndb/rest-server2/server/test_go/internal/mgmclient/mgmclient.go
+++ b/storage/ndb/rest-server2/server/test_go/internal/mgmclient/mgmclient.go
@@ -34,24 +34,41 @@
 //	result: Ok\n            (or an error message)
 //	...
 //	\n
+//
+// "get user" additionally returns "num_rows: N" and, after the blank line,
+// N rows of the form "key = value"; see GetUser.
 package mgmclient
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"net"
+	"strconv"
 	"strings"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


smkniazi:
The mgmd will retry a busy SET USER for up to 120s, but the Go test client gives up at 60s. So the client can report "failed" while the mgmd goes on to apply the change a minute later — tests
  then act on a rate limit they think didn't change.

smkniazi (approved):
The mgmd will retry a busy SET USER for up to 120s, but the Go test client gives up at 60s. So the client can report "failed" while the mgmd goes on to apply the change a minute later — tests then act on a rate limit they think didn't change.

PR #1088, opened 2026-09-14
RONDB-1092: by-pointer invalidateTable/invalidateIndex of a parked di…

mronstro:
…ctionary object (fixes RONDB-1121 F20)

Since RONDB-1092 (#1013) a stale local dictionary entry is parked when getTable()/getIndex() finds it Invalid or Altered, and a newer incarnation is cached under the same internal name. The by-pointer NdbDictionary::Dictionary::invalidateTable(const Table *) / invalidateIndex(const Index *) / removeCachedTable(const Table *) were still implemented as "drop the local entry by name, release once": after parking that destroyed the NEWER incarnation's wrapper (leaking its reference) and released the PARKED reference, which releaseStaleTableReferences() released again at the next request boundary. The count fell below the real holders and the object was deleted under another Ndb's live pointer.

RDRS hit it whenever RonSQL's stale-schema unload (invalidateIndex / invalidateTable by pointer) followed a parking, which the RDRS TTL watcher's by-name invalidate of a freshly created table triggers: a SIGSEGV in NdbDictionaryImpl::getIndex reading m_index from a freed object, or an abort in GlobalDictCache::release on a garbage name from releaseStaleTableReferences (RONDB-1121 F20, intermittent on every ronsql_fs MTR test after CREATE TABLE).

Fix: NdbDictionaryImpl::detach_local_reference releases the reference that backs the given pointer - the local entry only if it still points at that object, otherwise the parked entry (unpark_stale_object) - exactly once, and never touches a different incarnation cached under the same name; a pointer this Ndb no longer references is a no-op. Null guards: getIndex no longer dereferences a null m_index; park_stale_object tolerates an absent entry.

Regression: testDict -n InvalidateParkedByPointer (two Ndbs: park via a by-name invalidate on the second, invalidate the parked pointer on the first; the current entries must survive and the second Ndb keep working, for a table and an ordered index). Verified: testDict OK; ronsql_fs, ronsql_fs_jit, ronsql_fs_ng2r2 green three times with no RDRS crash.

Claude-Session: https://claude.ai/code/session_01Hqzb4DtLC8diLxnUfADW3N (cherry picked from commit 76cc05701c6d9e8aea502d7ef0c564fafcf70054)


smkniazi:
The test is not integrated in the mtr test.

smkniazi (approved):
The test is not integrated in the mtr test.

PR #1089, opened 2026-09-14
node_start_observability: uniform [NODE-START] logging of the data no…

KernelMaker:
…de start

A data node start passes through many phases spread over many kernel blocks, and until now each block wrote its own lines in its own style. A reader could not tell how far a start had come, how much was left, or whether a quiet stretch meant hard work or a stall; some waits printed nothing at all for minutes. This commit adds one uniform narrative of the start to the node log, tagged [NODE-START], so that a single grep tells the whole story:

  [NODE-START] plan: node restart (this node rejoins, keeping its
               data), executing 15/16 steps (skipping: 4 redo-init)
  [NODE-START] step 8/16 (restore) started
  [NODE-START] step 8/16 sub-step 2/2 (restore: restore fragments
               from LCP) progress: LDM(5): restored 5/6 assigned
               fragments, 469696 row operations so far, ...
  [NODE-START] step 8/16 (restore) completed: all 8 LDMs, elapsed=3s
  [NODE-START] completed: node restart finished, total elapsed=18s

The model
---------
Every start is described as the same 16 numbered steps: init, join, admission, redo-init, start-perm, redo-prepare, metadata, restore, undo-dd, redo-exec, index-rebuild, synchronize, wait-lcp, activate, handover, barrier. The numbering is the same for every start type, so step 10 is always the REDO execution. A plan line printed at admission says which steps this start will run, and the steps that do not apply are marked 'skipped' at the point where they would have run, so every log counts 1 to 16 exactly once. Steps have sub-steps where a phase of the step runs on its own (the metadata step has four in a node restart and two in a system restart, for example).

Each line carries a verb: started, progress, waiting, completed, skipped, failed. A progress line's numbers move (MBytes, files, fragments, pages, table objects, rows, with a rate and an estimate where rows are counted); a waiting line states the condition that is not met yet, names who is awaited, and repeats unchanged until it is met. The elapsed field of a step line counts from the start of the step (on a per-LDM line from that LDM's start of it) and that of a sub-step line from the start of the sub-step, so a completion gives the duration of the thing it completes. The steps that run inside the LDM threads (4, 6, 8, 9, 10, 11) print one completion per LDM and one node-wide line at the existing fan-in in the DBLQH proxy: 'all 8 LDMs', 'all 4 LDMs holding the 4 REDO log parts', 'N of M LDMs held fragments to restore', or 'no fragments to restore on this node'.

Progress and waiting lines are printed by a timer (one delayed CONTINUEB per active step per block instance), not by the work itself, so a wedged step keeps printing its unchanged counters and the stall is visible without any tooling. The period is the new configuration parameter NodeStartLogReportFrequency (seconds, default 15, 0 = the boundary lines only; the tick delay is capped at the time-queue limit and re-armed). Work a node does for another node's start is marked '[assist node N]', and boundary lines name the master, so the master's log shows the copy of the metadata, the invalidation of the old checkpoints and the checkpoint it drives for the starting node.

Where the lines go
------------------
The node log gets everything. The cluster log receives, through infoEvent, the plan line and the node-level started, completed and skipped lines of the steps (except step 1's started line, printed before the node can send events), and once a step has run for 60 s the report tick's line for it, repeated every 10 minutes, so a real stall reaches the cluster log without flooding it. A start that ends early prints exactly one '[NODE-START] failed:' line, from ErrorReporter, the watchdog's fast exit, the early ndbd_run exits (with the asynchronous log drained first), the angel's give-up exits or the forked child's exec failure; it is node-log only, since it is written by a process that is already going down.

What the long steps report
--------------------------
- step 1: the memory touch in MBytes, attributed to the init sub-step that runs it (with LateAlloc the pools are mapped when the blocks read their configuration).
- step 4: the REDO log file initialisation in MBytes and files per LDM.
- step 6: the REDO log file and MByte each part is reading, or a waiting line while a file open or close is in flight.
- step 7: on a system restart the sysfile synchronisation, then the schema restore with DBDICT's pass and schema object and the table objects read and distributed; on a node restart the master's four sub-steps (pause LCP, copy distribution information, copy dictionary information, include the node in the protocols) and the starting node's view of them. The counts are DICT table objects (tables, unique index tables and ordered indexes).
- steps 8 and 12: fragments and row operations with a rate; the RESTORE and recover-thread instances publish their per-fragment counts to the requesting DBLQH worker in batches, the take-over sums the rows received by the workers. Step 12's REDO-logging sub-step names the local checkpoint barrier it waits behind (N/M LDMs done, the GCI wait, the log tail cut) and counts the fragments of logged tables only, saying that the ordered indexes and unlogged tables need none.
- step 9: LGMAN's UNDO log head, pages and records applied, the page cache flush, TSMAN's extent scan.
- step 10: the round, part, file and position, and which LDMs have no log part or have finished the round.
- step 11: indexes and rows scanned with a rate and an estimate, from an atomic row count kept by DBTUX's builders and the fragment fan-out kept by DBTUP.
- step 13: whether the master's LCP is running with or without this node, following the master's classification of the LCP status.
- step 15: the subscribers still to connect, the switchover GCI and the nodes still to confirm it.
- the cluster-wide waits of an initial or system restart (start permission, the NDB start phase 4 wait point, the start phase barriers) as step-less 'waiting:' lines from NDBCNTR, and the system-restart master's wait for the other nodes' local recovery.

Blocks touched
--------------
QMGR, NDBCNTR, DBDIH, DBLQH and its proxy, DBDICT, DBTUP, DBTUX, RESTORE, LGMAN, TSMAN, SUMA, ErrorReporter, the ndbd and angel start-up code, ndbd_malloc, and the configuration parameter plumbing. The shared definitions (steps, sub-steps, applicability per start type, line formatting, the report timer) live in the new header NodeStartLog.hpp, whose header comment is the contract.

Non-logging changes are limited to Dblqh::c_master_node_id being maintained outside ERROR_INSERT builds, an optional retry callback in ConfigRetriever::allocNodeId, atomics in the memory-touch loop and the DBTUX row counter, release/acquire publication of the few DBDIH values the DBLQH proxy reads from the rep thread, and
Configuration::setLockCPU/addThread gaining a no-abort variant for the log thread's CPU binding. No recovery control flow changes.

Verification
------------
Every start type (initial start, system restart, node restart, initial node restart, a node without a node group, a node that cannot find its peers, starts that fail early and late) was run on two- and three-node clusters, with report frequencies 0, 1, 3, 15 and 60, small and large schemas (up to 1800 tables and 300 foreign keys), disk data, concurrent writes, and error inserts that stretch the individual steps. A checker read every resulting log against the contract: every executed step has exactly one started and one completed line, sub-steps start and complete in order, progress numbers move and their totals hold, elapsed matches the wall clock, the plan matches the skipped markers, and nothing follows the closing line. Every number the lines print was cross-checked against ndbinfo, the configuration or the legacy lines for the same work.


storage/ndb/src/kernel/blocks/dbdih/DbdihMain.cpp
```diff
@@ -1944,6 +1993,438 @@ I N T E R N A L  P H A S E S
...
+  const Dbdih *dih = (const Dbdih *)globalData.getBlock(DBDIH);
+  return (dih != nullptr) ? dih->nsl_sr_metadata_start() : 0;
+}
+
+bool nsl_dih_performed_copy_phase() {
+  const Dbdih *dih = (const Dbdih *)globalData.getBlock(DBDIH);
+  return (dih != nullptr) && dih->nsl_performed_copy_phase();
+}
+
+bool nsl_dih_wait_lcp_reported() {
```
mronstro, line 2009: Function never used

storage/ndb/src/kernel/blocks/dbdih/DbdihMain.cpp
```diff
@@ -1944,6 +1993,438 @@ I N T E R N A L  P H A S E S
...
 /*---------------------------------------------------------------------------*/
+/* Declared in NodeStartLog.hpp. NDBCNTR shares DBDIH's main thread; the
+   DBLQH proxy runs in the rep thread and gets the values DBDIH publishes
+   atomically (Dbdih::nsl_sr_*). */
+Uint64 nsl_dih_sr_metadata_start() {
+  const Dbdih *dih = (const Dbdih *)globalData.getBlock(DBDIH);
+  return (dih != nullptr) ? dih->nsl_sr_metadata_start() : 0;
+}
+
+bool nsl_dih_performed_copy_phase() {
```
mronstro, line 2004: Function never used

storage/ndb/src/kernel/blocks/dbdih/DbdihMain.cpp
```diff
@@ -1944,6 +1993,438 @@ I N T E R N A L  P H A S E S
...
+bool nsl_dih_wait_lcp_reported() {
+  const Dbdih *dih = (const Dbdih *)globalData.getBlock(DBDIH);
+  return (dih != nullptr) && dih->nsl_wait_lcp_reported();
+}
+
+bool nsl_dih_sr_receiving_tables(Uint64 &sub_start, Uint32 &tables) {
+  const Dbdih *dih = (const Dbdih *)globalData.getBlock(DBDIH);
+  sub_start = 0;
+  tables = 0;
+  if (dih == nullptr) {
```
mronstro, line 2018: dih == nullptr is an impossible thing

storage/ndb/src/kernel/vm/NodeStartLog.hpp
```diff
@@ -0,0 +1,814 @@
...
+  static Uint32 clamp(Uint32 pos, Uint32 len) {
+    return (pos >= len) ? (len - 1) : pos;
+  }
+
+  /* Shared body of line()/assist_line(); assist_node 0 = own start. */
+  static const char *vline(char *buf, Uint32 len, Uint32 assist_node,
+                           Uint32 step, Uint32 sub, Uint32 startType,
+                           const char *verb, Int64 elapsed_sec,
+                           const char *detail_fmt, va_list ap)
+      ATTRIBUTE_FORMAT(printf, 9, 0);
```
mronstro, line 532: RonDB and NDB is standardised on prints using g_eventLogger->info with timestamps in microseconds. Not sure how this fits into this model, simply doesn't is my analysis. Same comment on all other usage of printf that might be there.

storage/ndb/src/kernel/vm/NodeStartLog.hpp
```diff
@@ -0,0 +1,814 @@
...
+};
+
+inline const char *NodeStartLog::line(char *buf, Uint32 len, Uint32 step,
+                                      Uint32 sub, Uint32 startType,
+                                      const char *verb, Int64 elapsed_sec,
+                                      const char *detail_fmt, ...) {
+  va_list ap;
+  va_start(ap, detail_fmt);
+  const char *ret = vline(buf, len, 0, step, sub, startType, verb,
+                          elapsed_sec, detail_fmt, ap);
```
mronstro, line 542: I also have some difficulties finding the format string defining how the print actually happens.

storage/ndb/src/kernel/blocks/dbdih/DbdihMain.cpp
mronstro: Ignore

storage/ndb/src/kernel/blocks/dbdih/DbdihMain.cpp
mronstro: Ignore

storage/ndb/src/kernel/blocks/ndbcntr/NdbcntrMain.cpp
```diff
@@ -3323,19 +3459,24 @@ void Ndbcntr::waitpoint61Lab(Signal *signal) {
     cnoWaitrep6++;
     if (cnoWaitrep6 == cnoStartNodes) {
       jam();
+      nsl_unpark();
       NodeReceiverGroup rg(NDBCNTR, c_start.m_starting);
       rg.m_nodes.clear(getOwnNodeId());
       signal->theData[0] = getOwnNodeId();
       signal->theData[1] = CntrWaitRep::ZWAITPOINT_6_2;
       sendSignal(rg, GSN_CNTR_WAITREP, signal, 2, JBB);
```
mronstro, line 3467: Is it intentional not to call nsl_park here

storage/ndb/src/kernel/blocks/ndbcntr/NdbcntrMain.cpp
```diff
@@ -3369,19 +3510,24 @@ void Ndbcntr::waitpoint71Lab(Signal *signal) {
     cnoWaitrep7++;
     if (cnoWaitrep7 == cnoStartNodes) {
       jam();
+      nsl_unpark();
       NodeReceiverGroup rg(NDBCNTR, c_start.m_starting);
       rg.m_nodes.clear(getOwnNodeId());
       signal->theData[0] = getOwnNodeId();
       signal->theData[1] = CntrWaitRep::ZWAITPOINT_7_2;
       sendSignal(rg, GSN_CNTR_WAITREP, signal, 2, JBB);
```
mronstro, line 3518: Is it intentional not to call nsl_park here?

storage/ndb/src/kernel/blocks/ndbcntr/NdbcntrMain.cpp
```diff
@@ -3632,6 +3786,14 @@ void Ndbcntr::handle_start_phase_110(Signal *signal) {
       " until all restarting nodes have completed their recovery");
   infoEvent("Waiting at restart barrier for all restarting nodes");
 
+  c_nsl_barrier_timer.start_step();
+  {
+    char buf[NodeStartLog::BUF_SIZE];
+    infoEvent("%s", NodeStartLog::line(buf, sizeof(buf),
```
mronstro, line 3792: Now I get even more confused, NodeStartLog calls printf, but here it is used as input to infoEvent??

storage/ndb/src/kernel/blocks/ndbcntr/NdbcntrMain.cpp
```diff
@@ -3761,9 +3923,98 @@ void Ndbcntr::leave_restart_barrier(Signal *signal, const char *reason) {
...
+                         "this node is the master, %u of %u nodes have not"
+                         " yet completed NDB start phase 7",
+                         cnoStartNodes - cnoWaitrep7, cnoStartNodes);
+      break;
+    default:
+      return;
+  }
+  if (c_nsl_park_timer.escalate_due()) {
+    jam();
+    infoEvent("%s", buf);
```
mronstro, line 4014: I somehow get the feeling you use infoEvent quite often, are u sure that not many of these, or even all, should be g_eventLogger->info and go to the Node log. Cluster log should not be swamped with printouts of logs of recovery, those should end up in node log.

storage/ndb/src/kernel/blocks/ndbcntr/NdbcntrMain.cpp
```diff
@@ -7636,6 +7984,54 @@ void Ndbcntr::sendWAIT_ALL_COMPLETE_LCP_CONF(Signal *signal) {
...
+    ldms_done = ldms - m_outstanding_wait_lcp;
+  }
+  return 1;
+}
+
+/* Declared in NodeStartLog.hpp; DBDIH reads it in the same thread. */
+Uint32 nsl_cntr_local_lcp_barrier(Uint32 &ldms_done, Uint32 &ldms,
+                                  Uint32 &gci_needed, Uint32 &gci_done) {
+  const Ndbcntr *cntr = (const Ndbcntr *)globalData.getBlock(NDBCNTR);
+  if (cntr == nullptr) {
```
mronstro, line 8009: cntr will never be nullptr

storage/ndb/src/kernel/vm/NodeStartLog.hpp
KernelMaker: It doesn't use printf anywhere, this is just the format-checking attribute. `ATTRIBUTE_FORMAT(printf, 9, 0)` tells gcc that parameter 9 (`detail_fmt`) is a printf-style format so it can check the `%u`/`%s` against the arguments at the call sites; the `0` is because this variant takes a `va_list`. It's the same attribute `Logger::info()` and `SimulatedBlock::infoEvent()` are declared with.

The actual output is the standard one: `vline()` snprintf's the pieces into a stack buffer and then does `g_eventLogger->info("%s", buf)` (end of `vline()`, ~line 599), so the timestamps etc. are exactly as for every other line. The buffer is returned so that the call sites which are supposed to mirror the line to the cluster log can pass the same text to `infoEvent()`.

I'll add a comment above these declarations saying that, it's clearly easy to misread.

storage/ndb/src/kernel/vm/NodeStartLog.hpp
KernelMaker: There is no single format string, the line is assembled piece by piece in `vline()` just below: the `[NODE-START] ` prefix, an optional `[assist node N] `, `step X/16 `, an optional `sub-step M/T (step: sub-step name) `, the verb, then the caller's detail via vsnprintf, and finally `, elapsed=Ns` if an elapsed time was given. That's why parts can be present or absent per call. The header comment at the top of the file shows the resulting line shapes.

storage/ndb/src/kernel/blocks/ndbcntr/NdbcntrMain.cpp
KernelMaker: No printf involved (see my reply in NodeStartLog.hpp). `NodeStartLog::line()` prints the line to the node log with `g_eventLogger->info` and returns the buffer; the call site here then sends the same text to the cluster log with `infoEvent()`. So one line in each log, not a duplicate in either. It's the same dual-sink pattern as the `NR Status:` lines in DIH. Which lines get mirrored this way is the policy question you raise below at line 4014.

storage/ndb/src/kernel/blocks/dbdih/DbdihMain.cpp
KernelMaker: You're right, all blocks are created in `SimBlockList::load()` before any signal runs, so this can't be null. I'll drop the check here and in the three helpers above it (they have the same pattern).

storage/ndb/src/kernel/blocks/ndbcntr/NdbcntrMain.cpp
KernelMaker: Yes, intentional. This branch is the release: the master gets into `waitpoint61Lab` once per event (its own `ph6BLab` plus one `CNTR_WAITREP` per non-master), and this is the entry where the last node has reported, so there is nothing left to wait for. It unparks at 3462 and broadcasts 6_2.

The master parks in the `else if` right below (3471) the first time it comes here with nodes still missing; the `c_nsl_park != NSL_PARK_MASTER_6_1` guard is only there so later reports don't restart the timer. Non-masters park at 3479 after sending 6_1 and unpark when 6_2 arrives in `execCNTR_WAITREP`.

storage/ndb/src/kernel/blocks/ndbcntr/NdbcntrMain.cpp
KernelMaker: Same as for wait point 6.1 above: this is the release once every node has reported 7.1. The master parks in the `else if` at 3522, the non-masters at 3530, and both unpark on 7_2 in `execCNTR_WAITREP`.

storage/ndb/src/kernel/blocks/ndbcntr/NdbcntrMain.cpp
KernelMaker: Agreed, and done in 1a7db6a8c6f. The cluster log now only gets three kinds of `[NODE-START]` line: the plan line, the final `completed: <start type> finished` line, and the report-tick line of a step that has been running for 60 s (then once every 10 min). All the per-step started/completed/skipped lines are node-log only now — 59 infoEvent mirrors removed, 17 remain. A healthy start adds 2 lines to the cluster log instead of ~30, and the existing `Start phase N completed` lines are still there as the coarse markers; a start that stalls still shows up there after a minute.

The same commit adds the total to step 12's copy progress (`copied X/Y fragments from live nodes` with rate and estimate; the take-over's first pass over the fragments counts them), and takes care of your other two comments (the null checks, and a note in the header that `ATTRIBUTE_FORMAT` is only the compile-time format check).

storage/ndb/src/kernel/blocks/dbdih/DbdihMain.cpp
```diff
@@ -8129,8 +8124,7 @@ void Dbdih::nr_start_fragments(Signal *signal, TakeOverRecordPtr takeOverPtr) {
       return;
     }  // if
     ptrAss(tabPtr, tabRecord);
-    if (tabPtr.p->tabStatus != TabRecord::TS_ACTIVE ||
-        tabPtr.p->tabStorage != TabRecord::ST_NORMAL) {
+    if (tabPtr.p->tabStatus != TabRecord::TS_ACTIVE) {
```
mronstro, line 8127: This looks weird, looks to me like a logical change. Any explanation to this change?

KernelMaker: No change to what gets restored, only to which fragments the loop looks at. The `tabStorage != ST_NORMAL` test did not go away, it moved down to line 8161, right in front of `nr_start_fragment()`: START_FRAGREQ still goes out for exactly the same fragments (our replica of a logged table), with the same real-time break after it. For a logged table the two versions do the same thing; for an ordered index or an unlogged table the old code skipped the whole table with `continue`, the new code walks its fragments but only counts them.

The count is `c_nsl_frags_total`, the denominator of step 12's `copied X/Y fragments` progress line. The copy phase (`startNextCopyFragment`, line 8982) visits every active table including indexes, so the total has to include their fragments too, and this loop is the one place that already walks every fragment with real-time breaks before the copy starts. So line 8127 now uses the same table filter as line 8982, and the count matches what the copy will do.

mronstro (approved):

